PR #3, opened 2022-02-15
Create DR000: use github files and adr process

bettymw:
We have chose the ADR process because it is good.


decision-record/DR000-use-github-and-the-adr-process
```diff
@@ -0,0 +1,20 @@
+# DR000: Recording our project decisions on GitHub Files using the ADR Process
+
+Date: 2022-02-10
+
+## Status
+
+Accepted
```
0atman, line 7: This looks good!

0atman (approved):

PR #7, opened 2022-02-18
Propose ADR004 - Store form schemas in an API

danielburnley:


ADR/ADR004-store-form-schemas-in-an-api.md
```diff
@@ -0,0 +1,31 @@
...
+  Runner[Forms runner]
+  Designer[Form designer]
+  Manager --> API
+  Designer --> API
+  API --> Runner
+  Runner --> API
+```
+## Consequences
+
+> both positive and negative consequences of the decision
```
0atman, line 31: I think the concequences are both scalability and cost.

We can ramp up the public side of the service, the runner and api, to meet demand, and then scale it back dramatically (1000x maybe, even) to save running costs.

danielburnley: Yeah trade off and costs are probably one of the big things to consider - I'm not sure the cost/scalability trade offs of say, the designer/manager portion vs the runner - wonder if one ends up being cheaper to run if it was separate vs as a monolith

ADR/ADR004-store-form-schemas-in-an-api.md
```diff
@@ -0,0 +1,143 @@
...
+
+  managerService <--> formDatastore
+  formDatastore --> formRenderer
+```
+
+- Form manager and runner share the same DB
+- No need to build/maintain APIs
+- Shared DB schema is a concern
+
+### Option 4 - Monolithic manager and runner
```
danielburnley, line 119: In terms of development ease, a monolith is often the easiest one - We'd only have one service to deploy, one code base, dont need to manage dependencies between them etc etc. 

There's an argument I think to keeping a monolith but building it in a way that it's theoretically easy to progress towards separation - it'll be a lot easier and our volume is likely to be low enough for the first while that we could do that. But I wonder if that's creating problems down the line with anything unavoidable that could occur (I'm thinking of scale mainly)

0atman: I agree that a monolith is easiest to develop quickly. Though, I really like the idea of the runner and api being separate and horizontally scalable, it seems an inevitable requirement?

ADR/ADR004-store-form-schemas-in-an-api.md
```diff
@@ -0,0 +1,144 @@
...
+  formStore --> formsApi
+  runnerRenderer --> formsApi
+```
+
+- Stores the forms in a datastore with the API being a separate application
+- Runner depends on API
+- Three applications to deploy with dependencies, too complex?
+- Could cache API responses for speed
+
+### Option 2a - Runner as an API
```
danielburnley, line 39: A concern with these two (2a/2b) i think would be around the potential for the publishing/unpublishing workflow to end up causing a mis-match if messages get lost on the way to publish/unpublish. Causing the datastore for the manager to be out of sync with what's in the runner. 

You could mitigate this with by doing lookups on the runners published API, but noting the complication

ADR/ADR004-store-form-schemas-in-an-api.md
```diff
@@ -0,0 +1,144 @@
...
+Proposed
+
+## Context
+
+- Manager/Designer components don't need to be as scalable as the runner
+- Runner needs to scale to number of users
+  - Sustainability, runner not needed during quiet times
+- "Published forms" need to be accessible to the runner
+
+### Option 1 - Forms API
```
danielburnley, line 16: Wonder how scaling would impact the data store, sharing a datastore between Manager and Runner (via the API) means that the datastore itself will written to/read from by both at the same time, but the runner itself only has to read. Though I suppose caching could mitigate this somewhat - especially if we're just reading published schemas from the runner we could add caching to the API (cache invalidation being the most fun thing)

ADR/ADR004-store-form-schemas-in-an-api.md
danielburnley: yeah I agree it feels like an inevitable requirement - normally I'd err on the side of "you can just scale the monolith that's fine" but with our platform, assuming we're successful, the runner aspect is going to face a lot more traffic than the management side and it feels inevitable we'd want to scale those applications differently

thomasiles:
II like it! Some quick thoughts:

I hate microservices as much as anyone but I think separating the builder from
the runner makes sense. I think for security, scaleability and reliability
reasons it would be nice keep the runner isolated from the builder, neither
component being able to call the other.

I like the idea of less things to run/maintain and sticking files in buckets
would d give us a lot of reliability. I'm a bit nervous about not having any
way to manipulate forms between the builder and schema storage and between
schema storage and runner. An API gives us a place to make changes in the
future, covering things like:

- We'll probably need to be able answer questions about what a form looked like
  at a certain date

- If we are keeping the form in a JSON structure (or other file format - XML)
  etc. we will probably want to validate it before we put it in a DB/KV/S3 type
  store, I think if we don't things will get messy and we run the risk of
  having a bucket mixed with valid/invalid JSON

- We will probably need to be able to list the current live forms

- We might need to perform other tasks at the time of publish - convert the
  form into a PDF, notify people there is a new form being published

- We might need to 'provision' the runner for a new form or organisation. When
  the runner receives `GET /runner/forms/apply-juggling-license` does it lookup
  `s3://form-runner/apply-juggling-license.json` and serves the form if it
  finds it and returns a 404 if it's not there? If so, have we reduced the
  threat of the bad guy publishing an official looking government form to the
  ability of someone slipping in json files to our bucket. If it doesn't work
  like that do we need to tell the loader when we have a new user/form - e.g. a
  whitelist of forms, maybe set up a new domain etc. Without an API, can we do
  that easily

- will we ever replace the form runner with something different? Will we ever
  need two different form runners running at the same time which work
  differently - maybe a form runner for internal teams or during migration to a
  new technology. Will we ever change our mind about having a single form
  runner - in the future we might shard forms or have a Gov.uk Pay runner and a
  normal runner. Do we need more permissions for a runner 'getting' a form than S3 provides?

- Will we ever want to have more than one form builder? Maybe we rewrite it in
  rust in the future? Maybe we have different form builders running at the same
  time. Do we want all the above publishing logic in the builder?

- Publishing a new form whilst a form-filler-inner is in the middle of a
  journey might cause issues. We might need the runner to access multiple
  versions of a form at a time to prevent issues. We might get away with
  buckets and versioning for this but we might need some extra redirection.
  This might not matter at first but save and return may require it.

- We might not serve the schema as a whole - e.g. the runner reads pages rather 
  than whole forms schemas.

The 'publishing' side of these issues might be solved by putting more features
in the builder. I think this could be a good place to use something like Lambda
in addition to S3 to provide an API or publishing 'pipeline'?

0atman (approved):
The only thing that makes me nervous about this is synchronising the forms schema between the draft forms in the `builder` and the published forms in the `api`.

I think on balance that separating them out is good for scaling, for all the reasons given in this PR. But schema synchronisation hmmmmmmmmmmm.

thomasiles (approved):

ADR/ADR004-store-form-schemas-in-an-api.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+- Manager/Designer components don't need to be as scalable as the runner
+- Runner needs to scale to number of users
+  - Sustainability, runner not needed during quiet times
```
aliuk2012, line 13:
```suggestion
  - Sustainability, builder not needed during quiet times
```
Should this not be talking about forms-admin?

PR #8, opened 2022-02-18
Propose ADR005 - Use shared form runner

danielburnley:


ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Consequences
+
+We will be able to deploy updates to the runner more easily.
+
+We will be able to scale the service to the total number of people filling in forms, rather than having idle servers for every low use form, reducing energy usage.
+
+If the runner service encounters an error, all forms will be impacted rather than just one.
+
+Forms will have to be hosted on the same domain, with the path denoting which form to run.
```
DavidBiddle, line 78: This is pushing at the limits of my infrastructure knowledge a bit, but is this strictly the case? Wouldn't we be able to have multiple subdomains pointing to the same PaaS app, with the app deciding which form to show?

danielburnley: That might be possible actually, I wasn't 100% - I know in the runner you could get the hostname for the request, but I'm not sure how we'd programatically assign the domains for PaaS. Feels like it could be possible though

0atman (approved):
Good content, good decision

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+
+  Manager --> RunnerA
+  Manager --> RunnerB
+  Manager --> RunnerC
+  Manager --> RunnerD
+```
+
+In this structure, each form has its own runner which serves each form to the end user. This:
+
+- Limit the impact if one form runner encounters an error
```
james-sheppard-gov-uk, line 49: Limits

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
+# ADR005: Use multi-tenant runner
```
james-sheppard-gov-uk, line 1: Would "shared" be more appropriate than "multi-tenant"?

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+The forms platform has a "runner", the service that runs the forms itself to the user who is filling them in.
+
+There are two different ways to approach the runner - single or multi tenant. Two existing forms creation services (MOJ Forms, XGovFormBuilder) both use the single tenant approach.
```
james-sheppard-gov-uk, line 13: If we change the language, this would become "runner per form or shared runner"

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Single tenant structure
```
james-sheppard-gov-uk, line 15: Or "Runner per form"

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+  Manager --> RunnerD
+```
+
+In this structure, each form has its own runner which serves each form to the end user. This:
+
+- Limit the impact if one form runner encounters an error
+- Makes it more complicated to roll out updates to each runner as we scale
+- Requires us to run a server consistantly for all forms regardless of traffic, using up money + energy when not required
+
+### Multi tenant structure
```
james-sheppard-gov-uk, line 53: or "shared form"

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+    D[Form D]
+  end
+  Manager --> Runner
+```
+
+In this structure, there is one runner service that serves forms to the end user. This:
+
+- Enables us to have easier control over updating the runner where required (dependency version increases, service updates, etc.)
+- Enables the ability to scale to the overall load on the forms service, removing the need to run idle servers for forms that have no traffic
+- Makes errors in the forms runner service have wider impact as all forms will be effected rather than just the one on the runner encountering the error
```
james-sheppard-gov-uk, line 71: Increases the risk of a problem with one form affecting the performance of others

ADR/ADR005-use-multi-tenant-runner.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Decision
+
+We have decided to use a multi tenant runner to host and run multiple forms.
```
james-sheppard-gov-uk, line 75: or shared runner

james-sheppard-gov-uk (commented):
A few minor suggestions on language used - let's discuss

danielburnley:
👍🏻 to these suggestions, they make it a little cleaner and i think Shared runner/runner per form is way more clear

ADR/ADR005-use-shared-runner.md
```diff
@@ -0,0 +1,83 @@
...
+  Manager --> RunnerB
+  Manager --> RunnerC
+  Manager --> RunnerD
+```
+
+In this structure, each form has its own runner which serves each form to the end user. This:
+
+- Limits the impact if one form runner encounters an error
+- Makes it more complicated to roll out updates to each runner as we scale
+- Requires us to run a server consistantly for all forms regardless of traffic, using up money + energy when not required
```
DavidBiddle, line 51: Spelling: should be `consistently`

DavidBiddle (approved):

PR #10, opened 2022-02-18
ADR003-use-govuk-signon

james-sheppard-gov-uk:


DavidBiddle (approved):

ADR/ADR003-use-govuk-signon.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+GOV.UK Forms needs a way of authenticating users who will create and manage forms.
+
+It would be preferably to make use of an existing authentication system rather than create a new one. To explore the options, 2 prototypes were created to better understand the effort to onboard and the account creation process.
```
alice-carr, line 13: `preferably` -> `preferable`

alice-carr (approved):
Noticed one small typo but otherwise looks good!

ADR/ADR003-use-govuk-signon.md
0atman, line 13: We should write all our ADRs in simple language .When this is edited, could you switch it all to first person, rather than the mix of first and third that we currently have?

For example, rewrite _"To explore the options, 2 prototypes were created..."_ to _"We created 2 prototypes to explore the options..."_

Third person is archaic and inevitably complicates what you are trying to say. Be direct: If we did it, say that. :+1:

alice-carr: Isn't `To explore the options, 2 prototypes were created..` technically the passive voice and not necessarily third person?

Either way we should avoid the passive voice too as active is much clearer

0atman: Passive voice! Yes! That's what I meant. It's even worse than third person, because you're not even NAMING the person who did the action :sweat_smile:

DavidBiddle: Agree that it would make sense to use the active voice here (even though I habitually use the passive voice all the time)

ADR/ADR003-use-govuk-signon.md
```diff
@@ -10,7 +10,7 @@ Proposed
 
 GOV.UK Forms needs a way of authenticating users who will create and manage forms.
 
-It would be preferably to make use of an existing authentication system rather than create a new one. To explore the options, 2 prototypes were created to better understand the effort to onboard and the account creation process.
+It would be preferable to make use of an existing authentication system rather than create a new one. We created 2 prototypes to better understand the effort to onboard and the account creation process.
```
0atman, line 13: the commenter was satisfied

PR #17, opened 2022-03-07
Add alpha & beta architecture diagram

danielburnley:
# What

Adds both the alpha and beta architecture diagrams into the repo using mermaid.

# Why

Github wiki does not currently allow mermaid to be used for creating diagrams within them, to get around this I've created the `diagrams/` folder and placed both the alpha architecture diagram as it was and the beta one we can evolve going forward within. This will let us refer to it in the wiki as we want, as well as keep it up to date without needing to rely on images


0atman (approved):
these diagrams are fantastic!

0atman:
@james-sheppard-gov-uk lets merge this in for now and iterate

james-sheppard-gov-uk:
Thanks for doing this @danielburnley !

As discussed this morning, I have a few of suggestions:
- let's start with a single file rather than 2, and make use of version history to refer back when needed
- I don't think the "001" prefix is needed in the filename, how about simply `architecture.md` (it's in the diagrams folder after all)

There's some styling improvements to be made, but let's do them next / later. I'd also like to keep our users on the diagram if we can, but again that's something to look at another time.

0atman:
@james-sheppard-gov-uk I agree our ARCHITECTURE.md should be the current version, but I see no harm in having an alpha one too, if it's clearly marked? Agree about no need for 001 prefix

PR #23, opened 2022-04-19
Revision: give us option to charge high vol services

harryvos:
In more recent team discussions following some cost modelling, our tech costs would be much higher than forecast if higher volume services started to use GOV.UK Forms. This revision would allow us to charge for these forms in the future if that meant the majority of forms can be created and run for free.


DavidBiddle (approved):
Looks good to me 🎉

PR #24, opened 2022-04-19
Revision: remove goals + dates from public repo

harryvos:
These aren't set in stone, and are rough guides for our team. They're probably better discussed internally


DavidBiddle (approved):
Good thinking!

PR #28, opened 2022-05-20
ADRXX - Form data structure

danielburnley:


diagrams/initial-api-schema-proposal.json
```diff
@@ -0,0 +1,100 @@
...
+      "url": "https://api.server.test/v1"
+    }
+  ],
+  "paths": {},
+  "components": {
+    "schemas": {
+      "form": {
+        "type": "object",
+        "properties": {
+          "id": {
```
thomasiles, line 18: Do we need a version in Form as well as the openapi version? It might make it easier to iterate if the runner is able to serve multiple versions of the form structure at the same time. It might also make it more complicated and isn't something we need yet.

thomasiles, line 18: Do we need a version in Form as well as the openapi version? It might make it easier to iterate if the runner is able to serve multiple versions of the form structure at the same time. It might also make it more complicated and isn't something we need yet.

diagrams/initial-api-schema-proposal.json
```diff
@@ -0,0 +1,100 @@
...
+              "$ref": "#/components/schemas/page"
+            }
+          }
+        }
+      },
+      "page": {
+        "type": "object",
+        "properties": {
+          "position": {
+            "type": "integer",
```
thomasiles, line 45: Could an approach more like a linked list than an array serve us better? I like the approach of the submit prototype - each page has a next field, which can be a lisp like cond. If we restrict his to value to only linking to questions ahead of the current page and not including the the current page, it solves a lot of issues like recursive form-filler journeys.

https://github.com/alphagov/submit-prototype-kit/blob/master/examples/simple.json

thomasiles, line 45: Could an approach more like a linked list than an array serve us better? I like the approach of the submit prototype - each page has a next field, which can be a lisp like cond. If we restrict his to value to only linking to questions ahead of the current page and not including the the current page, it solves a lot of issues like recursive form-filler journeys.

https://github.com/alphagov/submit-prototype-kit/blob/master/examples/simple.json

diagrams/initial-api-schema-proposal.json
```diff
@@ -0,0 +1,100 @@
...
+          },
+          "questionType": {
+            "description": "The type of question being asked",
+            "allOf": [
+              "$ref": "#/components/schemas/questionTypes"
+            ]
+          }
+        }
+      },
+      "translatableText": {
```
thomasiles, line 74: Really like this approach - simple and robust.

diagrams/initial-api-schema-proposal.json
```diff
@@ -0,0 +1,100 @@
...
+          "cy": {
+            "type": "string",
+            "description": "The text in welsh"
+          }
+        }
+      },
+      "questionTypes": {
+        "type": "string",
+        "enum": [
+          "singleLine",
```
thomasiles, line 90: Do we need page type - for the message when a user has submitted the form or some other way of adding it

danielburnley: I've added these in as static properties on the form object itself

diagrams/initial-api-schema-proposal.json
```diff
@@ -0,0 +1,100 @@
...
+            "description": "The hint text for the question",
+            "allOf": [
+              "$ref": "#/components/schemas/translatableText"
+            ]
+          },
+          "questionType": {
+            "description": "The type of question being asked",
+            "allOf": [
+              "$ref": "#/components/schemas/questionTypes"
+            ]
```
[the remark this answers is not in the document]
danielburnley: Yeah this makes sense - `/<form>/page/<id>` make sense. I can imagine it'll be a string auto-generated from the title rather than a numerical ID though for the runner for the URLs to be readable to analytics platforms

diagrams/initial-api-schema-proposal.json
danielburnley: Yeah i've been pondering this - I'm thinking we could do something like this for now where it's a simple page ID that's next  (how we generate this will be its own challenge, i think the Admin UI and this representation will end up being some interesting glue) and when we investigate something more in depth later we can change the `next` value to take something more, maybe a simple condition like in submit.

danielburnley:
@alphagov/forms-devs 

Given something @0atman said monday talking through this (to paraphrase, why are we doing schema first rather than code first) - I'm beginning to feel like this ADR itself likely shouldn't get merged and we should close the PR, and follow it up with a at least one or two specific PRs about certain bits that explain our reasoning but allowing the code to drive the less notable bits. 

The two off the top of my head are:

### Translatable text from day 1

The idea that we're going to present all text to the runner in "translatable text" objects so that we can add them in the future more easily, even if the Admin side doesn't support that feature for a while it's worth building it in,

### How we handle "next page" logic

@thomasiles brought up a good point [here](https://github.com/alphagov/forms/pull/28#discussion_r879570099) looking at the submit example. The XGov Form builder does the same, though i can't find for sure what MoJ forms does. But it feels like a reasonable approach to go with to begin with - especially for giving us room to adapt. As @thomasiles pointed out, it gives us room to add additional criteria in the future.

----

Given the above, unless anyone disagrees, i'll likely make another look over this tomorrow and then close it after any final comments, looking to create the two above as they're more specific in scope and probably more valuable to discuss.

0atman:
Agreed. Another thing is that with `grape-oapi` we can generate the schema from the code.
If we're either building a schema in 1. a swagger builder, or 2. ruby, I choose the code!

danielburnley:
Closing for the reasons mentioned [above](https://github.com/alphagov/forms/pull/28#issuecomment-1137468845)

I'll follow this up with two additional ADRs, one for translatable text and the other for page routing

PR #30, opened 2022-05-26
ADR007: Add translatable text ADR

danielburnley:


DavidBiddle (approved):
Seems like a sensible approch to me, and very clear explanation in the ADR 👍

PR #32, opened 2022-05-31
ADR008 - Page ordering defined by pointers to the next page in form objects

danielburnley:


ADR/ADR008-page-ordering-defined-by-pointers.md
```diff
@@ -0,0 +1,84 @@
...
+Of the three options only option 3 gives the ability to extend it in the future for branching purposes as both 1 and 2 are based on a linear ordering whereas 3 gives us the opportunity to expand it in the future, by replacing the string value with an object which can provide the rules for branching.
+
+## Decision
+
+The team has opted to go with option 3.
+
+## Consequences
+
+- We will be able to navigate from one page to the next
+- Navigating backwards will be more challenging as we don't have a pointer backwards
```
DavidBiddle, line 83: How do we reckon we'll tackle this? 
1. keep track of the previous pages in a stack?
2. work it out from the answers the user has given?
3. adding `previous` to these page objects as well as `next`? i.e. make this a doubly linked list?
We don't necessarily have to decide this now but if we're going to do option 3 it might make sense to include it in this ADR.

DavidBiddle: That makes sense to me

harryvos:
Thanks for sharing these three options in the ADR. Would it be worth adding a short description of the impact on migration to and from GOV.UK Forms by us going with "Option 3: Treat pages as a linked list, with a pointer to the next page"?

I'm thinking about potential scenarios where colleagues may want to import one or a bunch of forms to GOV.UK Forms from other GOV.UK form building platforms (and vice versa) without having to create them from scratch. I couldn't find the specific schemas, but as two examples:

https://github.com/XGovFormBuilder/digital-form-builder/blob/main/model/src/schema/schema.ts
https://ministryofjustice.github.io/form-builder-metadata-api-docs/#tag/Services/operation/latest-service-metadata

Alignment with existing schemas might help to converge in the future?

harryvos:
@danielburnley Good question. Thanks for breaking the challenge down a bit more.

We've had a few emails from people looking to migrate to GOV.UK Forms, mostly from COTS platforms, and two from government platforms, though they've not specifically asked for "import existing forms" as a feature.

I was more thinking about generally trying to converge and avoid lock-in within UK central government platforms (COTS vendors generally aren't incentivised to do this). I think this ADR would benefit from documenting where we think using an existing approach would cause problems for us, both in the schemas and the other aspects you mentioned? I'd hope that would be useful for future developers on the team to understand why we are taking our own approach.

jenbutongit:
👋 I know that this is long closed.. but opt 2 (point to next page vs position) would be better. Opt 1 won't work when you have significant branching. I haven't yet implemented this for our runner, but the graph view in our designer is based on computational graphs. you can use algorithms to go forwards and backwards. and typically very well documented.. since it's just maths. 

singly linked should be ok. I'd love to work a solution for this that works for all of us!

PR #36, opened 2022-08-04
Create DR007-form-is-live-first-then-add-draft.MD

harryvos:
# PR Checklist

---

# What

New decision record.

# How to review

Please see if there's enough useful information in the decision record without duplicating everything in the Google Doc.

# Who can review

Anyone in the GOV.UK Forms team except me.

### DIFF
--- a/decision-record/DR007-form-is-live-first-then-add-draft.MD
+++ b/decision-record/DR007-form-is-live-first-then-add-draft.MD
@@ -1,0 +1,45 @@
+# DR007: We'll get ready to support a live form first, then build draft when we have time
+
+Date: 2022-08-01
+
+## Status
+
+Proposed
+
+## Context
+
+We need to know what is needed for the first form, so we know what work we need to prioritise. At the moment, in the development version of GOV.UK Forms, there is only one concept of a live form. There is no separate "draft" or preview version. There is no versioning. The way we would build the platform to support the first form would likely differ, depending on how we want to support the form creators at the Insolvency Service to make changes to their first live form. What happens to an end user and their data if they're in the middle of completing the form when the form creator makes a change?
+
+[More context for team members](https://docs.google.com/document/d/1xPmYDPYmZXCBkvWIV6QoDuGLaS-iSEq1jk6aehIfz6k/edit#)
+
+## Decision
+
+We'll finish getting everything ready to support a live form first (Option 1), then build a way to create and edit a draft when we have time (Option 4).
+
+### Option 1: All forms created with GOV.UK Forms are “live”. Forms are edited directly
+
+```mermaid
+graph TD
+    B(Live) --> |Edit form| B
+```
+
+### Option 4: A single “draft” version of a form can be worked on while a separate “live” version exists
+
+```mermaid
+graph TD
+    A[New version] --> B(Draft)
+    B --> |Edit form| B
+    B --> |Publish form| D(Live)
+    D --> |Create new draft - Live form stays up| A
+```
+
+## Consequences
+
+While GOV.UK Forms will soon have a way to draft forms, if we don't have this in time for our partner to make their first form live, we're comfortable that the associated risks are low enough, particularly around disrupting end users if a change is made to a live form.
+
+We have discarded these options:
+
+- Option 2: Each form starts as “draft”, then can be made “live” but can’t be set to “draft” again. “Live” forms are edited directly
+- Option 3: Each form starts as “draft”, then can be made “live”. “Live” forms can’t be edited but can move back into “draft”, making form unavailable until it is “live” again
+
+as for the extra effort, they don't offer much benefit to our users, as changes to live forms are still disruptive, as with Option 1.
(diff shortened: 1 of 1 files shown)


decision-record/DR007-form-is-live-first-then-add-draft.MD
```diff
@@ -0,0 +1,45 @@
...
+
+Date: 2022-08-01
+
+## Status
+
+Proposed
+
+## Context
+
+We need to know what is needed for the first form, so we know what work we need to prioritise. At the moment, in the development version of GOV.UK Forms, there is only one concept of a live form. There is no separate "draft" or preview version. There is no versioning. The way we would build the platform to support the first form would likely differ, depending on how we want to support the form creators at the Insolvency Service to make changes to their first live form. What happens to an end user and their data if they're in the middle of completing the form when the form creator makes a change?
```
deberny, line 11: We need to agree on what option we will provide to our form creators to make their first form go live, so that we can prioritise the work for the upcoming sprints.

In the current development version of the GOV.UK Forms form-runner, there is no concept of a draft. So forms that are created and edited in the form-admin are displayed in preview on form-runner as 'live'.

This is not the intended technical implementation, as we'd like to be able to support 'daft' and 'live' state of a form, with potential for versioning. It is however a first pass option that will let us make the first form live quicker.

We are aware that the first (current) option will have an impact on the end user if a form creator makes direct edits to a live form and can potentially disrupt a form filling journey mid way. For example: What happens to an end user and their data if they're in the middle of completing the form when the form creator makes a change?

We are thinking that this is unlikely to happen for the first form, but we're will be looking into making that eventuality less disruptive.

We will be striving to design, test and build GOV.UK Forms to be more closely aligned with the option 4, where "A single “draft” version of a form can be worked on while a separate “live” version exists".

deberny: @james-sheppard-gov-uk  is mentioning form-runner and admin too much?

james-sheppard-gov-uk: @deberny I suggest we keep the technical details out, just focus on the user experience.

deberny: We've had a chat and agreed that technical bits are not needed. So I'll take my comment back.

decision-record/DR007-form-is-live-first-then-add-draft.MD
```diff
@@ -0,0 +1,45 @@
...
+    A[New version] --> B(Draft)
+    B --> |Edit form| B
+    B --> |Publish form| D(Live)
+    D --> |Create new draft - Live form stays up| A
+```
+
+## Consequences
+
+While GOV.UK Forms will soon have a way to draft forms, if we don't have this in time for our partner to make their first form live, we're comfortable that the associated risks are low enough, particularly around disrupting end users if a change is made to a live form. If a change is made to a live form while an end user is filling in the form, they will probably get an error page, and have to start again. If multiple changes are made in a row, this could be frustrating for the end user.
+
```
james-sheppard-gov-uk, line 39: I think it would be worth capturing the consequences for our partners, i.e. they need to be made aware of the risks / limitations of working without drafts; when draft functionality is added this needs to be communicated effectively to whoever has already started creating and editing forms so they understand the changes.

james-sheppard-gov-uk (commented):
I've added a comment to the ## Consequences section

james-sheppard-gov-uk (changes requested):
We need to include the consequences for our partners, i.e. they need to be made aware of the risks / limitations of working without drafts; when draft functionality is added this needs to be communicated effectively to whoever has already started creating and editing forms so they understand the changes.

decision-record/DR007-form-is-live-first-then-add-draft.MD
DavidBiddle: Am I right in thinking that this is now resolved, since we (sort of) had draft functionality prior to the first form being made?

decision-record/DR007-form-is-live-first-then-add-draft.MD
```diff
@@ -0,0 +1,50 @@
...
+## Context
+
+We need to know what is needed for the first form, so we know what work we need to prioritise. At the moment, in the development version of GOV.UK Forms, there is only one concept of a live form. There is no separate "draft" or preview version. There is no versioning. The way we would build the platform to support the first form would likely differ, depending on how we want to support the form creators at the Insolvency Service to make changes to their first live form. What happens to an end user and their data if they're in the middle of completing the form when the form creator makes a change?
+
+[More context for team members](https://docs.google.com/document/d/1xPmYDPYmZXCBkvWIV6QoDuGLaS-iSEq1jk6aehIfz6k/edit#)
+
+## Decision
+
+We'll finish getting everything ready to support a live form first, then add Option 2 on to allow private editing of forms.
```
james-sheppard-gov-uk, line 17: Suggest "editing" is changed to "creation"

decision-record/DR007-form-is-live-first-then-add-draft.MD
james-sheppard-gov-uk: My initial comment related to Option 1. One of the consequences of Option 2 is that the forms-admin now contains a warning when making a form live
> Any changes you make to a live form will be updated in the form immediately.

james-sheppard-gov-uk (approved):

PR #51, opened 2022-12-01
Add ADR for use of AWS CodePipeline

danworth:
ADR for the use of CodePipeline and CodeBuild for managing deployments to the new GOV.UK Forms AWS Environments.


DavidBiddle (approved):
Sounds like a very sensible approach to me 🎉

PR #55, opened 2023-01-10
Supersede Notify ADR

0atman:
Discuss the problems here, and we'll add them to this ADR, and make a new ADR with our decision

### DIFF
--- a/ADR/ADR002-use-govuk-notify.md
+++ b/ADR/ADR002-use-govuk-notify.md
@@ -4,7 +4,7 @@ Date: 2022-04-07
 
 ## Status
 
-Rejected
+Superseded
 
 ## Context
 
--- a/ADR/ADR002-use-govuk-notify.md
+++ b/ADR/ADR002-use-govuk-notify.md
@@ -4,7 +4,7 @@ Date: 2022-04-07
 
 ## Status
 
-Accepted
+Rejected
 
 ## Context
 
--- a/ADR/ADR002-use-govuk-notify.md
+++ b/ADR/ADR002-use-govuk-notify.md
@@ -4,7 +4,7 @@ Date: 2022-04-07
 
 ## Status
 
-Superseded
+Superseded by [ADR017 Use Amazon SES](ADR017-use-amazon-ses.md)
 
 ## Context
 
--- a/ADR/ADR002-use-govuk-notify.md
+++ b/ADR/ADR002-use-govuk-notify.md
@@ -4,7 +4,7 @@ Date: 2022-04-07
 
 ## Status
 
-Superseded by [ADR017 Use Amazon SES](ADR017-use-amazon-ses.md)
+Superseded by [ADR018 Use Amazon SES](ADR018-use-amazon-ses.md)
 
 ## Context
 
--- a/ADR/ADR018-use-amazon-ses.md
+++ b/ADR/ADR018-use-amazon-ses.md
@@ -1,4 +1,4 @@
-# ADR017: Use Amazon SES
+# ADR018: Use Amazon SES
 
 Date: 2023-03-27
 
--- a/ADR/ADR018-use-amazon-ses.md
+++ b/ADR/ADR018-use-amazon-ses.md
@@ -36,6 +36,10 @@ There isn't currently a way to audit which [GOV.UK Notify API keys](https://docs
 
 ### Implementation
 
+Amazon Simple Email Service (SES) addresses the Attachments, Formatting and Security issues outlined above.
+
+GOV.UK Forms is in the process of being migrated onto AWS, so using Amazon SES is the obvious choice.
+
 GOV.UK Notify uses Amazon SES to send emails, so there is already knowledge and experience within GDS on how to integrate with it.
 
 ## Decision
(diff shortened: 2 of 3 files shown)


ADR/ADR002-use-govuk-notify.md
```diff
@@ -4,7 +4,7 @@ Date: 2022-04-07
 
 ## Status
 
-Accepted
+Rejected
```
DavidBiddle, line 7: Think this would be superseded rather than rejected? 
```suggestion
Superseded
```

0atman: perf

0atman:
@james-sheppard-gov-uk agreed with @DavidBiddle, saying
> you're absolutely right that we should have an ADR about changing the way we send emails. However, we should supersede the existing one
> If a decision is reversed, we will keep the old one around, but mark it as superseded. (It's still relevant to know that it was the decision, but is no longer the decision.)
https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions

0atman:
Reading the ADR and adding the TLS problem to it suggests there are 4 problems with using Notify:


1. GOV.UK Notify does not support email attachments
2. GOV.UK Notify does not support adding reply-to email addresses over the API
3. Data sent in the emails will be stored for 7 days
4. GOV.UK Notify uses 'best effort' TLS for end-to-end secure email delivery.

0atman:
My initial take on these, though I feel out of the loop and really would like to be corrected is:

1. This is a feature and a deliberate design decision we should honour.
2. `no-reply` is the standard method of contacting your users
3.  Is this a problem because the emails contain PII? Not using attachments and instead providing a link to log in and view the data could solve this. Email is not a secure database, we should not fool ourselves that it is.
4. This is only a problem if our form creators use non government emails. We have an allowlist of valid domains for log in, this is impossible.

aliuk2012:
> 2. GOV.UK Notify does not support adding reply-to email addresses over the API

Not strictly true. It is possible to add/change the reply-to but the reply to address does need to be manually added via the Notify UI to get an `id` for the api to use. https://docs.notifications.service.gov.uk/ruby.html#email-reply-to-id-optional

DavidBiddle:
There's also:
5. Notify's Markdown implementation has some undocumented features, and this was one of the factors that contributed to the issues with decimals we saw in October. 
6. The Notify team don't really want us using their platform for this - they don't really see us as the kind of transactional service they're meant to support.

DavidBiddle (dismissed):
This looks good and I think is a fair representation of the previous conversations we've had about this.

@james-sheppard-gov-uk is there a backlog card for switching over to SES? If not I feel like we should probably make one, since once we merge this we're sort of committed to doing it.

ADR/ADR018-use-amazon-ses.md
```diff
@@ -0,0 +1,51 @@
...
+
+There isn't currently a way to audit which [GOV.UK Notify API keys](https://docs.notifications.service.gov.uk/rest-api.html#api-keys) have been used, and what for. This limits protective monitoring.
+
+### Implementation
+
+GOV.UK Notify uses Amazon SES to send emails, so there is already knowledge and experience within GDS on how to integrate with it.
+
+## Decision
+
+Use [Amazon SES](https://aws.amazon.com/ses/) directly instead of GOV.UK Notify for sending emails to form processing teams.
```
cadmiumcat, line 43: Wondering if we've done a spike on this or on whether there are other services out there that would work? I haven't been following closely so sorry if this has already been discussed.

If that's already done it might be worth linking to it or listing the reasons why we're choosing Amazon SES over others?

ADR/ADR018-use-amazon-ses.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Consequences
+
+* emails sent to processing teams will go via Amazon SES. This includes:
+    1. verification emails to confirm the correct email address
+    2. submitted forms
+* A `from` email address needs to be chosen and DNS needs to be configured securely
+* GOV.UK Forms doesn't yet send email confirmations to members of the public who have completed forms. If / when this functionality is added, we need to decide if GOV.UK Notify is the most appropriate
```
lfdebrux, line 51: Is this line saying we might consider using Notify as well as SES for emails? There are a couple of other use cases I can think of, for instance if we start emailing form creator users about user management things, we could use Notify for that.

ADR/ADR018-use-amazon-ses.md
james-sheppard-gov-uk: No spikes done, SES is the obvious choice: we're on AWS so aligns with the rest of our infrastructure; Notify uses SES so plenty of learning there.

ADR/ADR018-use-amazon-ses.md
james-sheppard-gov-uk: Yes, we'll consider what to use when the need arises.

james-sheppard-gov-uk:
Thanks for the discussion earlier - I've summarised the choice of Amazon SES under "Context, Implementation".

lfdebrux (approved):

DavidBiddle (approved):

DavidBiddle (approved):

PR #61, opened 2023-02-24
Update making live sequence diagrams

aliuk2012:
Updating these two diagrams in prep for when we have draft/live forms

[Change a live form diagram](https://github.com/alphagov/forms/blob/436544ff16ed0df21d39ab3bd72018aaeabc295e/diagrams/sequence-diagrams/changing-a-form.md)

[Publishing a form diagram](https://github.com/alphagov/forms/blob/436544ff16ed0df21d39ab3bd72018aaeabc295e/diagrams/sequence-diagrams/publishing-a-form.md)

### DIFF
--- a/diagrams/sequence-diagrams/changing-a-form.md
+++ b/diagrams/sequence-diagrams/changing-a-form.md
@@ -1,6 +1,7 @@
 # Changing a live form
 
 ```mermaid 
+%%{init:{"sequence": { "showSequenceNumbers": true }}}%%
 sequenceDiagram
   title Changing a live form
   actor user as form creator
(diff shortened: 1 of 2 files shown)


diagrams/sequence-diagrams/publishing-a-form.md
```diff
@@ -15,22 +15,20 @@ sequenceDiagram
...
 
   user->>browser: Confirm making form live<br />click "Save and continue" button
   browser->>admin: POST /forms/{form id}/make-live
-  admin->>api: GET /forms/{form id}
-  admin->>api: GET /forms/{form id}
-  admin->>api: GET /forms/{form id}/pages
-  admin->>api: PUT /forms/{form id}/make-live
-  admin-->>browser: 302
+  admin->>api: GET /api/v1/forms/{form id}
+  admin->>api: GET /api/v1/forms/{form id}
```
DavidBiddle, line 25: Was this double request resolved by https://github.com/alphagov/forms-admin/pull/332 ?
```suggestion
  admin->>api: GET /api/v1/forms/{form id}
```

DavidBiddle (commented):
Looks good overall and matches my expectations of how this would work. Potentially you might want someone else who's worked on draft/live to take a look but will leave that up to you.

Have also raised a question about the double request from Admin to API that's happening a few times in the diagrams.

diagrams/sequence-diagrams/publishing-a-form.md
aliuk2012: No, this is the part where I noticed there were 3 requests being made (2 of those were triggered by using `current_form`. After that PR you have mentioned it was reduced to 2 calls. 

Basically all our requests make two calls to that endpoint, the first one is for authorization check to ensure the form belongs to the org, and the next request is for the form itself. I have a ticket to have a look at these double requests as part of the firebreak.

DavidBiddle: Ah OK - that makes sense, thanks!

DavidBiddle (approved):

diagrams/sequence-diagrams/changing-a-form.md
```diff
@@ -9,25 +10,54 @@ sequenceDiagram
...
+    admin-->>browser: REDIRECT 302
+    browser->>admin: GET /forms/{form id}
+    admin->>api: GET /forms/{form id}
+    admin->>api: GET /forms/{form id}
+    admin->>api: GET /forms/{form id}/pages
+    browser-->>user: show draft form page with task list
+  else form creator decides to make form live
+    user->>browser: click "Yes" and submits page
+    browser->>admin: POST /forms/{form id}/make-live<br>payload: {forms_make_live_form"=><br>{"confirm_make_live"=>"made_live"},<br>"form_id"=>"{form id"}}
+    admin->>api: POST /api/v1/forms/{form id}/make-live<br>payload: {includes all form attributes and values}<br>Controller Action doesn't use any
```
aliuk2012, line 50: This endpoint eventually uses this model method to make a form live, as you can see we don't use any of the form attributes as forms-api sets the `live_at` using its own internal clock.

aliuk2012: https://github.com/alphagov/forms-api/blob/76f048e9d919d9a150109e5d9d543f24411ca364/app/models/form.rb#L9-L11

PR #62, opened 2023-02-27
ADR: Adding PaperTrail to forms-api for auditing/restoration of our ActiveRecords.

aliuk2012:
[ADR for adding PaperTrail ](https://github.com/alphagov/forms/blob/adr-paper-trail-for-auditting/ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md)

PR to add PaperTrail to forms-api: https://github.com/alphagov/forms-api/pull/212

### DIFF
--- a/ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
+++ b/ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
@@ -35,7 +35,8 @@ come to do User Management, we can set up the extra details as part of that work
 ### Positives
 
 - We can identify exactly what has changed and when it was changed.
-- We can restore from a specific version of Form/Page
+- We can restore from a specific version of Form/Page. This can be done through migration or Rails console, as a team we also
+  discussed the option of building an endpoint/UI which super admins may have access to to do these sort of tasks.
 - PaperTrail adds call back into all the crud actions create/update/delete so its easy to track those events
 - PaperTrail also accepts custom events an can be manually triggered
 - It's really easy to tell PaperTrail to track a model by including `has :papertrail` to the model that you would like to
--- a/ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
+++ b/ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
@@ -21,8 +21,9 @@ changes to Forms/Pages but also from any other tables we might want to monitor.
 
 ## Consequences
 
-We will not be able to identify what has changed and when without having to go back to splunk server logs and trawl 
-through them in the hopes of being able to spot the changes. We also need this information to surface back to the user.
+We will be able to identify very easily what attribute on a specific model changed, when it changed and who changed it.
+This means that we do not have to trawl splunk server logs trying to see what information had been sent to the server
+and at what point did the model get updated to. We also need this information to surface back to the user.
 
 By setting it up now, we can double check the data that's being collected and tweak it before surfacing it to our users. 
 If there are any issues we can reconfig or completely drop and restart this data collection without any questions from our users.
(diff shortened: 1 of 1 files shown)


ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
```diff
@@ -0,0 +1,51 @@
...
+If there are any issues we can reconfig or completely drop and restart this data collection without any questions from our users.
+
+We already know that we will need to do some additional work to identify who made the changes as we don't record that now as 
+forms-api does not have this data. We will need to pass it in from forms-admin. If we do setup PaperTrail now, when we 
+come to do User Management, we can set up the extra details as part of that work.
+
+### Positives
+
+- We can identify exactly what has changed and when it was changed.
+- We can restore from a specific version of Form/Page
```
danworth, line 37: Would this be done via an active record migration? If not what is the proposed method?

ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
```diff
@@ -0,0 +1,51 @@
...
+
+Accepted
+
+## Context
+
+We need way to audit changes to database records to identify what has changed and who made those changes. This information
+will be surfaced to users later on when we come to versioning. This is still valuable to record now for any support tickets
+that come in and require us to understand the form/pages history.
+
+There is also a secondary need be able to restore individual records rather than having to restore from database backups.
```
danworth, line 15: I haven't seen this need, do you have a link to a Trello or similar? Is this a requirement from versioning and offering the ability for admins to revert to an earlier version or as some kind of support activity by us? Links to my question below about how this would actually be done.

How does papertrail deal with versions for records that have been deleted? Does it keep their versions and can they be restored in the way you're intending?

ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Decision
+
+Add [PaperTrail](https://github.com/paper-trail-gem/paper_trail) as part of draft/live work so we can start recording 
+changes to Forms/Pages but also from any other tables we might want to monitor.
+
+## Consequences
+
+We will not be able to identify what has changed and when without having to go back to splunk server logs and trawl
```
danworth, line 24: This should be written form the perspective of the consequences of accepting this ADR (this reads as from the perspective or not accepting it and relying on app logging instead).

ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
aliuk2012: It could be a migration or we could use rails console to restore it.

As mentioned at this morning standup, we could create a UI/param endpoint that would rollback a version but until we have our first request we don't really know what method we will end up with. The point is that either way it is possible to restore it. I'll add some more content to the ADR about it

ADR/ADR013-use-paper-trail-gem-for-auditing-and-record-backups.md
aliuk2012: There isn't a specific Trello card for it. I am aware that there has been discussions around offering users the option to possibly rollback etc but its still very early days. I think for the time being this is kind a support activity/tool for us to be able to revert changes (if theres a large number of request we can look at either making it a super admin feature or perhaps allow specific users with the correct permissions to do it). What I was trying to do was point out that as part of capturing audit trails there is also a way to restore the records to their given state.

To answer you delete question, papertrail logs `deletes` as an event and as part of that loging it stores the state that the record was in when it was deleted. This means that we can restore the record by pulling out the delete event and using the state the record was last in to restore it. https://github.com/paper-trail-gem/paper_trail#3a-reverting-and-undeleting-a-model

danworth: Great thanks for the clarification Al. I don't wish to be pedantic but I wonder if this should be phrased as "There is also a secondary benefit where it could be possible to restore..." since "need" to me implies something concrete thats on the roadmap.

aliuk2012: No problem, suggestions more than help 😄 👍

thomasiles (approved):
looks good to me

PR #68, opened 2023-03-06
Move decision to top of template

DavidBiddle:
# PR Checklist

---

# What

I've moved the 'Decision' section above 'Context' in the ADR template. The decision is the crucial piece of information in an ADR, the context and consequences are supporting information. This is also closer to the [original ADR proposal by Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions), as referenced in the GDS Way. We briefly discussed this at today's technical catchup.

# How to review

Semantic: Do you agree with the changes?

# Who can review

Any Forms tech team member

### DIFF
--- a/ADR/ADRXXX-architecture-decision-record-template.md
+++ b/ADR/ADRXXX-architecture-decision-record-template.md
@@ -6,13 +6,13 @@ Date: YYYY-MM-DD
 
 > Proposed, Accepted, Rejected, Deprecated, Superseded by ADRzzz
 
-## Context
-
-> the facts behind the need to make the decision
-
 ## Decision
 
 > what the team has decided to do
+
+## Context
+
+> the facts behind the need to make the decision
 
 ## Consequences
 


lfdebrux (commented):
If we like this change, it might be worth flagging up with GDS Way peeps and seeing if there's interest in adopting this format more widely.

ADR/ADRXXX-architecture-decision-record-template.md
```diff
@@ -2,6 +2,10 @@
 
 Date: YYYY-MM-DD
 
+## Decision
+
+> what the team has decided to do
+
 ## Status
 
 > Proposed, Accepted, Rejected, Deprecated, Superseded by ADRzzz
```
lfdebrux, line 11: Personally I'd be tempted to put the status before the decision, otherwise I think it might get lost visually...

DavidBiddle: I debated doing this too - given that we're allowing `Status: Rejected`, I think it probably is important that this doesn't get lost, so will move status back to the top

james-sheppard-gov-uk:
Thanks for proposing this @DavidBiddle, although I don't agree with the change.

The summary in [The GDS Way - How to document decisions](https://gds-way.cloudapps.digital/standards/architecture-decisions.html#how-to-document-decisions) lists

* Title: a description of the decision (not the problem)
* Status: for example, Proposed, Accepted, or Superseded
* Context: the facts behind the need to make the decision
* Decision: what the team has decided to do
* Consequences: both positive and negative consequences of the decision

The order in the [original ADR proposal by Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions) is

* Title
* Context
* Decision
* Status
* Consequences

(I think that putting Status nearer the top make sense)

Both of these have the Decision following the Context. For me, the Title should summarise the decision. The examples Michael Nygard gives are

> "ADR 1: Deployment on Ruby on Rails 3.0.10" or "ADR 9: LDAP for Multitenant Integration"

If the reader wants to understand more, they can then choose to read the rest of the ADR in full, with the Context setting the scene for the more detailed Decision that was made.

Keen to hear other views!

DavidBiddle:
@james-sheppard-gov-uk I've no idea how I managed to miss the word CONTEXT at the top of Michael Nygard's post, but somehow I did 😬

lfdebrux:
> @james-sheppard-gov-uk I've no idea how I managed to miss the word CONTEXT at the top of Michael Nygard's post, but somehow I did 😬

To be fair, I also missed it 😅

DavidBiddle:
Closing this - think it probably isn't offering much that writing descriptive titles wouldn't!

PR #69, opened 2023-03-07
New ADR for Identity Provider Authentication and implementing authorisation

james-sheppard-gov-uk:
# PR Checklist

---

# What

I've created a new ADR to explain the move away from using GOV.UK Signon for authentication and authorisation.

# How to review

Do you agree with the ADR? Does it explain the decision well?

# Who can review

@alphagov/forms-devs

### DIFF
--- a/ADR/ADR014-idp-authentication.md
+++ b/ADR/ADR014-idp-authentication.md
@@ -1,10 +1,10 @@
-# ADR013: Use another Identity Provider for authentication
+# ADR014: Use an Identity Provider for authentication and implement authorisation
 
 Date: 2023-02-06
 
 ## Status
 
-Proposed
+Accepted
 
 ## Context
 
--- a/ADR/ADR014-idp-authentication.md
+++ b/ADR/ADR014-idp-authentication.md
@@ -12,7 +12,9 @@ In [ADR003 Use GOV.UK Signon](ADR003-use-govuk-signon.md) we recognised that
 
 > We don't yet know if/how self-service account creation could work
 
-Self-service account creation is essential to allow adoption to scale in public beta. This is not going to be possible with GOV.UK Signon so we need to find an alternative way to authenticate our users.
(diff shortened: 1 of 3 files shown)


ADR/ADR014-idp-authentication.md
```diff
@@ -0,0 +1,25 @@
...
+
+Accepted
+
+## Context
+
+In [ADR003 Use GOV.UK Signon](ADR003-use-govuk-signon.md) we recognised that
+
+> We don't yet know if/how self-service account creation could work
+
+Self-service account creation is essential to allow adoption to scale in public beta. This is not going to be possible with GOV.UK Signon so we need to find an alternative way to authenticate our users.
```
hannahkc, line 15: I'm not sure how much detail we include in these but is it worth saying very briefly why it's not possible to use Signon for self/service account creation? Is it because Signon accounts have to be created by a GDS admin?

ADR/ADR014-idp-authentication.md
```diff
@@ -0,0 +1,25 @@
+# ADR014: Use an Identity Provider for authentication and implement authorisation
+
+Date: 2023-02-06
+
+## Status
+
+Accepted
```
aliuk2012, line 7: Should this be superseded rather than Accepted because it changes our approach to how we authenticate users?

DavidBiddle: I think this one would be Accepted, and ADR003 should be changed to superseded once this is merged (this PR might be the right place to do that).

james-sheppard-gov-uk: @aliuk2012 @DavidBiddle I've superseded ADR003. The choice of an Identity Provider will in turn supersede this ADR.

ADR/ADR014-idp-authentication.md
james-sheppard-gov-uk: Thanks for your comments @hannahkc @bettymw - I've added some detail.

PR #72, opened 2023-03-17
Add ADR for switching to a Vite-based build

DavidBiddle:
# PR Checklist

---

# What

Adds an ADR for using Vite instead of our existing frontend build setup. It's also worth reading the [runner PR for doing this](https://github.com/alphagov/forms-runner/pull/281), which contains some of the same information, but also includes additional implementation details and test results specific to that repo.
 
# How to review

# Who can review

Anyone except me


ADR/ADRXXX-use-vite-rails-for-frontend-build.md
```diff
@@ -0,0 +1,41 @@
+# ADRxxx: Use Vite and the vite_rails gem for our frontend build
```
DavidBiddle, line 1: Will add a number before merging

lfdebrux:
I'm just trying to read up a bit on Vite/Vite Ruby; I've not used this particular tool before, it seems like a reasonable choice so far though. I wanted to ask though, where there any alternatives you considered? My impression is that frontend tooling is a pretty crowded market at the moment, so just wondering why you picked Vite particularly.

ADR/ADRXXX-use-vite-rails-for-frontend-build.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Decision
+
+We have decided to move from Gulp/Rollup/Sprockets and the Rails CSS/JSBundling gems to a build based on vite-rails.
+
+## Consequences
+
+- Faster frontend builds and updates - starting the dev server should be quicker, as should updates when code changes.
+- Faster builds overall - from local testing it appears that a clean install should be quicker, because we don't need to install as many gems or Yarn.
+- It's theoretically possible to for the dev server to do something different from the prod build - this is most likely to happen when devloping for browser-specific bugs. In these cases this can be circumvented by running a full build of the assets and then starting a local rails server, rather than using the vite dev server.
```
james-sheppard-gov-uk, line 41: typo?
> possible ~to~ for

DavidBiddle:
@lfdebrux Good point - I've now added some notes on the alternatives I considered. Ultimately I thought Vite and Parcel were the best fits for the project, because both offer a good balance between speed and convenience. Vite edged it for me because of the speedy local dev server and the fact it uses Rollup for bundling, so would be fairly familiar in that respect if we do need to maske any configuration changes.

lfdebrux:
> @lfdebrux Good point - I've now added some notes on the alternatives I considered. Ultimately I thought Vite and Parcel were the best fits for the project, because both offer a good balance between speed and convenience. Vite edged it for me because of the speedy local dev server and the fact it uses Rollup for bundling, so would be fairly familiar in that respect if we do need to maske any configuration changes.

Thanks for the detailed reply, this is very helpful and all sounds good to me :)

PR #82, opened 2023-05-22
Add ADR for using GOV.UK Organisation API

lfdebrux:
# What

Document the decision to use the GOV.UK organisation API. See [ADR018](./ADR/ADR018-use-govuk-organisation-api.md) and ticket https://trello.com/c/kKEiA0Bq for details.

# How to review

# Who can review

- Product manager
- Tech architect
- Developers
- Content designer
- Tech writer

### DIFF
--- a/ADR/ADR018-use-govuk-organisation-api.md
+++ b/ADR/ADR018-use-govuk-organisation-api.md
@@ -1,0 +1,63 @@
+# ADR018: Use the GOV.UK organisation API
+
+Date: 2023-05-22
+
+## Status
+
+Proposed
+
+## Context
+
+We are migrating from GOV.UK Signon to another identity provider (see [ADR015](./AD015-idp-authentication)). Currently when users sign in Signon sends us information about the organisation they belong to. Alternative identity providers we are considering will not necessarily be able to provide that organisation information, so we must manage and store it within our own service.
+
+The list of known organisations in GOV.UK Signon comes from the [GOV.UK organisations API]. This API serves the same list of government organisations as shown on the GOV.UK website at [gov.uk/government/organisations](https://www.gov.uk/government/organisations). The API also includes historical organisations, information about the organisation including its type and relationships with other organisations, and a stable and universally unique identifier for each organisation (the "content ID"). In total the GOV.UK organisation API currently lists 1178 historical and current government organisations, including central government departments, agencies and other public bodies. It does not include local government bodies, however for the public beta of GOV.UK Forms we are targeting central government and do not expect any users from local government.
+
+Using the GOV.UK organisation API will give us access to a ready made list of organisations for our users (including organisations for all existing users), and will mean we do not have to create any user interface for creating and managing organisations within our service. The main tradeoff is we introduce a dependency with another team within GDS.
+
+[GOV.UK organisations API]: https://docs.publishing.service.gov.uk/repos/collections/api.html
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


DavidBiddle (dismissed):
Looks good to me - clear and comprehensive run down of the reasons for the decision, and matches the discussions we've had on this previously.

cadmiumcat (approved):
👍 this makes sense

danworth:
I can see the benefit of this in that we keep the organisations to some official list. I can see that also presenting a limitation in  how services might be able to use Forms which may or may not be a problem. On Pay for example, we had multiple accounts from the same government service/council to represent how they saw things internally. I believe our authorisation is based upon Org (and the users role). So if multiple teams within say HMRC (for which there is one org listed but is a very large organisation) wanted to maintain separate forms/permissions based upon the internal structure of HMRC can that be done e.g. "HMRC team 1" and "HMRC team 2" or do we not see that as a user need?

lfdebrux:
> I can see the benefit of this in that we keep the organisations to some official list. I can see that also presenting a limitation in how services might be able to use Forms which may or may not be a problem. On Pay for example, we had multiple accounts from the same government service/council to represent how they saw things internally. I believe our authorisation is based upon Org (and the users role). So if multiple teams within say HMRC (for which there is one org listed but is a very large organisation) wanted to maintain separate forms/permissions based upon the internal structure of HMRC can that be done e.g. "HMRC team 1" and "HMRC team 2" or do we not see that as a user need?

Great point. We've discussed this with product peeps, we're currently thinking that having teams that are at sub-organisation level will be further down the line, maybe after public beta. It would definitely need a lot more work.

My main thinking from a technical point of view are that the approach in this ADR doesn't necessarily stop us from having organisations that aren't in the GOV.UK list in the database, or even in the same table; as long as we're making sure to keep the content ID reserved for organisations that are known to GOV.UK, and have some method for avoiding clashes with slugs. There will obviously be a great deal many other considerations for our models about how users are assigned to teams, how the hierarchy is expressed, etc., but as we don't support those things currently anyway I feel like it's fine to leave those considerations for later.

hannahkc:
Looks very clear and great to me. The only bit I have a comment in is this: 

> When users come to us asking for their organisation name to be changed, we will have to liaise with publishers who have access to Whitehall Admin. 

I think for government organisations, we ought to be asking them to go through their GOV⁠.⁠UK publishing team and GOV.UK content team to do this as normal - without our involvement. And then we would respond to the change in the relevant way once it's been made in the API? I think this is good as it means the API remains the source of truth and we don't interfere with the process for maintaining and updating it.

adamrobertson-gds:
Looks good to me, thanks!

lfdebrux:
> Looks very clear and great to me. The only bit I have a comment in is this:
> 
> > When users come to us asking for their organisation name to be changed, we will have to liaise with publishers who have access to Whitehall Admin.
> 
> I think for government organisations, we ought to be asking them to go through their GOV⁠.⁠UK publishing team and GOV.UK content team to do this as normal - without our involvement. And then we would respond to the change in the relevant way once it's been made in the API? I think this is good as it means the API remains the source of truth and we don't interfere with the process for maintaining and updating it.

Makes sense @hannahkc. If I update that line to read

> When users come to us asking for their organisation name to be changed, we will have to tell them to talk to publishers in their organisation who have access to Whitehall Admin or to the GOV.UK content team.

does that better express what you'd expect?

hannahkc (approved):

PR #89, opened 2023-06-14
Update version-1.md - minor edits

C-Harry:
Have made some small changes to wording for readability and house style. 

I have a few queries/comments but didn't want to put these within the text itself, so I'll try adding these as 'suggestions' once I've opened a PR request. Not sure if this is the best/only way to do this? Would be good to discuss for future reference.

# PR Checklist

---

### DIFF
--- a/design/features/routing/version-1.md
+++ b/design/features/routing/version-1.md
@@ -49,8 +49,6 @@ ___
   - the question the route takes the person to is deleted
   - the question the route takes the person to is now above the question the route should start from
 - We'll also show an error when a question the route takes the person to is now directly below the question the route starts from 
-  - This could make our database complicated and have an impact on future iterations
-  - This is mostly house keeping
 - We will not allow a form to be made 'Live' without the breaking change errors being resolved
   - This is to reduce the likelihood of poor forms, and infinite loops within forms
 - We will not allow a creator to change the question the route is based on when editing a route
--- a/design/features/routing/version-1.md
+++ b/design/features/routing/version-1.md
@@ -28,39 +28,39 @@ ___
 
 ### As-is
 
-- form creator creates questions and marks them as optional giving content as part of hint text to explain who needs complete a question
+- Form creator creates questions and marks them as optional. Form creators can use hint text to explain who needs to complete a question.
 
 ### To-be
 
-- from creator can select radio option and set a page to go to if selected. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
-- form creator HAS to select a next question for each radio option IF one radio option in a list has had a next question set  
-- form creator can set next (single) question as skippable if radio option X is chosen  
+- Form creator can create a 'question route' by selecting a radio option ('select from a list' question) and setting a page to go to depending on the answer to the route's start question. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
+- Form creator **must** select a next question for each radio option **if** one radio option in a list has had a next question set  
+- Form creator can set next (single) question as skippable if radio option X is chosen  
 
 
 ## Key decisions
 
-- we are only going to allow routing to be based on a radio button answer type
-- there can only be one route for a question
-  - a question can be routed to and then routed from
-  - but cannot have multiple routes from
-- we will not be considering branching journeys as part of this work
-- we will prioritise errors for breaking changes, such as
-  - the answer to base the route on is changed or deleted
+- We're only going to allow routing to be set up from a radio button answer type 
+- Only one route can be set up for each question
+  - A question can be routed to and then routed from
+  - A question cannot have multiple routes from it
+- We will not consider branching journeys as part of this work
+- We'll prioritise errors for breaking changes, such as:
+  - the answer the route is based on is changed or deleted
   - the question the route takes the person to is deleted
-  - the question the route takes the person to is now above the question the route should start
-- we will also give an error for when a question the route takes the person to is now directly below the question the route starts from 
-  - this could make our database complicated and have an impact on future iterations
-  - this is mostly house keeping
-- we will not allow a form to be made “Live” without the breaking change errors being resolved
-  - this is to reduce poor forms, and infinite loops within forms
-- we will not allow a creator to change the question the route is based on when editing a route
-  - the user should delete a route in this situation, as they will need to select all the different parts of the route
-  - this is something we might revisit, but makes development easier for now
-- moving a page that has a route attached (starting question) will also move the route
-- if the route start question is deleted, the route will also be deleted
-- if the route start question answer type is changed from radio options, the route will be deleted. This includes
-  - answer type being changed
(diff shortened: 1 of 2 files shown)


design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
-- form creator creates questions and marks them as optional giving content as part of hint text to explain who needs complete a question
+- Form creator creates questions and marks them as optional. Form creators can use hint text to explain who needs to complete a question.
 
 ### To-be
 
-- from creator can select radio option and set a page to go to if selected. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
-- form creator HAS to select a next question for each radio option IF one radio option in a list has had a next question set  
-- form creator can set next (single) question as skippable if radio option X is chosen  
+- Form creator can create a 'question route' by selecting a radio option ('select from a list' question) and setting a page to go to depending on the answer to the route's start question. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
+- Form creator **must** select a next question for each radio option **if** one radio option in a list has had a next question set
```
C-Harry, line 36: Does "a next question" refer to the destination question/page that someone's being routed to?

design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
-- there can only be one route for a question
-  - a question can be routed to and then routed from
-  - but cannot have multiple routes from
-- we will not be considering branching journeys as part of this work
-- we will prioritise errors for breaking changes, such as
-  - the answer to base the route on is changed or deleted
+- We're only going to allow routing to be set up from a radio button answer type 
+- Only one route can be set up for each question
+  - A question can be routed to and then routed from
+  - A question cannot have multiple routes from it
```
C-Harry, line 45: Is this bullet point not repeating the main bullet ("Only one route can be set up for each question")?

design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
-  - the user should delete a route in this situation, as they will need to select all the different parts of the route
-  - this is something we might revisit, but makes development easier for now
-- moving a page that has a route attached (starting question) will also move the route
-- if the route start question is deleted, the route will also be deleted
-- if the route start question answer type is changed from radio options, the route will be deleted. This includes
-  - answer type being changed
-  - option for “People can only select on option” is unselected on the “Create a list of options” screen, making the question a checkbox and not a radio type question
+  - the question the route takes the person to is now above the question the route should start from
+- We'll also show an error when a question the route takes the person to is now directly below the question the route starts from 
+  - This could make our database complicated and have an impact on future iterations
```
C-Harry, line 52: Not quite sure what this bit means (or what the 'This' refers to)

design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
+  - This is mostly house keeping
```
C-Harry, line 53: Not sure what the 'This' refers to again here

design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
 
-- form creator creates questions and marks them as optional giving content as part of hint text to explain who needs complete a question
+- Form creator creates questions and marks them as optional. Form creators can use hint text to explain who needs to complete a question.
 
 ### To-be
 
-- from creator can select radio option and set a page to go to if selected. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
-- form creator HAS to select a next question for each radio option IF one radio option in a list has had a next question set  
-- form creator can set next (single) question as skippable if radio option X is chosen  
+- Form creator can create a 'question route' by selecting a radio option ('select from a list' question) and setting a page to go to depending on the answer to the route's start question. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)
```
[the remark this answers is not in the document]
christophercameron-ixd: As discussed (21/06/2023) these bullets came from the Mural board and were very early on design agnostic ’to-be‘ criteria of where we should/could aim. 

The ‘set page’ in this is indeed the destination where the user would be routed to - and these ‘to-be’s were incase we had to custom set the page to take the user to for every option in the list.

design/features/routing/version-1.md
```diff
@@ -28,57 +28,57 @@ ___
...
+- Form creator creates questions and marks them as optional. Form creators can use hint text to explain who needs to complete a question.
 
 ### To-be
 
-- from creator can select radio option and set a page to go to if selected. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
-- form creator HAS to select a next question for each radio option IF one radio option in a list has had a next question set  
-- form creator can set next (single) question as skippable if radio option X is chosen  
+- Form creator can create a 'question route' by selecting a radio option ('select from a list' question) and setting a page to go to depending on the answer to the route's start question. If a radio option has no set page then option will go to next page in the order (unless another option has the next page set as the next step)  
+- Form creator **must** select a next question for each radio option **if** one radio option in a list has had a next question set  
+- Form creator can set next (single) question as skippable if radio option X is chosen
```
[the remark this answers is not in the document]
christophercameron-ixd: As per the above comment. Plus this was in a specific instance where the user might make the next question skippable if an answer is chosen - so the opposite way we’ve implemented. As we set the page to take the user to not the page to hide/skip.

design/features/routing/version-1.md
christophercameron-ixd: Yeah, it does I think I wrote it to clarify when it came to development as part of the conversations we were having. 
Happy to remove or keep

design/features/routing/version-1.md
christophercameron-ixd: The ‘this’ refers to the situation described in the above bullet point. It’s an extra level of detail to explain why we consider it to be an essential error to cover as part of the initial breaking changes error work we did

design/features/routing/version-1.md
christophercameron-ixd: As above

design/features/routing/version-1.md
C-Harry: I'm still not sure what it is that could make our database complicated exactly? (And what the relevance of this is here.) 

So are we saying that if someone moves a route's destination question so that it sits directly below the route's start question it not only makes the route unnecessary but also complicates our database somehow?

Could we be a bit more explicit about how it would 'complicate' the database?

Maybe say something like 'Besides making the route redundant, this could cause [bit more detail about how it would affect the database]'?

design/features/routing/version-1.md
C-Harry: I'm also not sure what this comment means.

C-Harry: Meaning this is housekeeping to keep the database in as clean and simple a state as possible..?

design/features/routing/version-1.md
christophercameron-ixd: Removed

PR #92, opened 2023-07-14
Add ADR for e2e tests moving to their own repo

SamJamCul:
# What

Created an ADR discussing separating the e2e tests from `forms-deploy` and moving them into a new repo called `forms-e2e-tests` 
# How to review

# Who can review
Anyone who's interested in the e2e tests!

### DIFF
--- a/ADR/ADR020-e2e-tests-get-own-repo.md
+++ b/ADR/ADR020-e2e-tests-get-own-repo.md
@@ -4,7 +4,7 @@ Date: 2023-07-14
 
 ## Status
 
-Proposed
+Accepted
 
 ## Context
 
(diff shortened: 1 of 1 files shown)


ADR/ADR020-e2e-tests-get-own-repo.md
```diff
@@ -0,0 +1,26 @@
...
+## Decision
+
+> what the team has decided to do
+
+## Consequences
+
+ - We have to (/get to) keep track of a new repo
+ - Tests can be updated without requiring the `forms-deploy` repo to be in order
+ - Test repo can have clearer dependencies and environment management
+
```
lfdebrux, line 24:
```suggestion
- End to end tests can be public (forms-deploy repo is private)

```

SamJamCul: Appended, good spot!

lfdebrux:
Presumably this will require some changes to our deploy pipelines, to get the end to end test code along with the app code and the terraform. I don't think this would cause issues though? 🤔

danworth (approved):
Subject to other team member opinions this seems like a good idea. Keep in mind that the pipelines will have to be updated to source the e2e from the new repo and the code build projects updated to run them. There is also a [script to run the end-to-end tests](https://github.com/alphagov/forms-deploy/blob/main/infra/scripts/end_to_end.sh) which will need to be updated (likely moved into the new repo under /bin ?). I'm happy to help in the planning regarding dependencies/approach and if you need any help from the infra team in doing so.

ADR/ADR020-e2e-tests-get-own-repo.md
```diff
@@ -0,0 +1,27 @@
...
+## Consequences
+
+ - We have to (/get to) keep track of a new repo
+ - Tests can be updated without requiring the `forms-deploy` repo to be in order
+ - Test repo can have clearer dependencies and environment management
+- End to end tests can be public (forms-deploy repo is private)
+
+
+> both positive and negative consequences of the decision
```
DavidBiddle, line 27: Think we can remove this line

SamJamCul: It's gone!

DavidBiddle (approved):
Looks good! 🎉

PR #98, opened 2023-08-10
Create ADR021: Use Auth0

james-sheppard-gov-uk:
# PR Checklist

---

# What

Created new ADR for using Auth0.

# How to review

1. Semantic: Do you agree with the ADR?

# Who can review

Anyone can review.

### DIFF
--- a/ADR/ADR021-use-auth0.md
+++ b/ADR/ADR021-use-auth0.md
@@ -32,6 +32,8 @@ We will use [Auth0 passwordless authentication with email](https://auth0.com/doc
 
 We make use of the [Auth0 OmniAuth Strategy gem](https://github.com/auth0/omniauth-auth0).
 
+Auth0 offers limited [customisation of the login pages](https://auth0.com/docs/customize/universal-login-pages/universal-login-page-templates), which may restrict use of the [GOV.UK Design System](https://design-system.service.gov.uk/).
+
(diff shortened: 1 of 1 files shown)


ADR/ADR021-use-auth0.md
```diff
@@ -0,0 +1,41 @@
...
+
+> Where possible, reduce reliance on passwords and implement passwordless authentication
+
+and there is [NCSC guidance on one time passwords](https://www.ncsc.gov.uk/guidance/authentication-methods-choosing-the-right-type#section_5).
+
+[UK Government Security Single Sign-On](https://sso.service.security.gov.uk/) was considered, but is still in an "Alpha" phase.
+
+[Auth0 offers Passwordless Authentication Methods](https://auth0.com/docs/authenticate/passwordless/authentication-methods) so was also considered, and a DPIA (Data Protection Impact Assessment) was carried out.
+
+[User research on authentication](../research/2023-08-Usability_Testing_Authentication.md) showed that participants were comfortable using a passwordless process and preferred the Auth0 journey.
```
jamie-o-wilkinson, line 23: Is it worth calling out that Auth0 can't be configured to use the design system exactly (from what I recall)? I don't think this was actually picked up on by any one in the user research but it was a concern we had beforehand.

james-sheppard-gov-uk: Thanks @jamie-o-wilkinson , good suggestion. I'll add something to the "Consequences" section, but if you're able to reference any specific concerns they could be included.

jamie-o-wilkinson (approved):

PR #102, opened 2023-09-14
Add ADR 22: Store forms in forms-admin before sending to forms-api

chao-xian:
Credit for a large portion of the Context and Strategy sections of this ADR to @thomasiles - thank you!

[Rendered view](https://github.com/alphagov/forms/blob/add-adr-22/ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md)

[Trello](https://trello.com/c/iHPh87YA/1031-rfc-for-switching-to-postgresql-in-forms-admin-instead-of-redis)

# PR Checklist

---

# What

Adds an ADR for forms-admin data persistence.

# How to review

Spike from Al

For example:

# Who can review

@alphagov/forms-devs

### DIFF
--- a/ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
+++ b/ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
@@ -1,0 +1,68 @@
+# ADR022: Store forms in forms-admin before sending to forms-api
+
+Date: 2023-09-14
+
+## Status
+
+> Accepted
+
+## Context
+
+See https://app.mural.co/t/gaap0347/m/gaap0347/1695033470183/a0f2dd9bdd9b698b1a411c8bfbcc33a9e5f987b0?sender=10492031-8763-46d6-ab0b-b41eb037cc4e
+
+Originally forms-admin didn't have a workflow but a single screen for the user to draft their form. This got broken up into a multistage workflow. Data is transferred between screens via cookies but this has limitations, and only worked ok for a linear flow. Now that we are approaching diverging workflows, we can't support this any longer. We are also currently only using ActiveModel for our classes.
+
+Users configure `pages` in forms admin over multiple steps (sometimes called a wizard 🧙!). `Pages` are only saved in a form when a user has completed all the required steps. Different answer types have difference steps.
+
+For example, the user chooses the "date" answer type, then on the following screen selects that it will be a date of birth and on the last screen enters the question text. The user then saves the `page`, at which point the `page` is added to the `form`.
+
+Currently we use the [form object pattern](https://dev.to/drbragg/rails-design-patterns-form-object-4d47#:~:text=Form%20Object%20Pattern,consistent%20API%20in%20our%20code.) to manage the intermediate state of pages as they are being created.
+
+ This gives each screen a [model](https://github.com/alphagov/forms-admin/tree/main/app/forms/pages), [controller](https://github.com/alphagov/forms-admin/tree/main/app/controllers/pages) and [view](https://github.com/alphagov/forms-admin/tree/main/app/views/pages).
+
+We store the state of incomplete `pages` in the session hash. The code updating this hash is spread across the form objects and controllers. When adding or modifying steps in the `page` journey, it's up to the programmer to make sure the session is updated correctly, such as making sure it's cleared in the right places.
+
+As we've added more features for `pages` such as extra input types, conditions and additional guidance, managing state in the session hash has become harder. We've had issues around producing inconsistent data, using a mix of strings and symbols and deleting data at the wrong point.
+
+Also, as the session hash is currently stored in user's cookie, there is a size limit of 4kb. This is not an immediate issue but could pose difficulties when a `page` has many question options, long guidance text or we start using the session in other parts of the application.
+
+## Strategy
+
+We should:
+
+This ADR proposes that we stop using the forms object pattern to create pages. Instead each step is represented using an activeRecord model and database table. This means we no longer need to use the session hash to maintain the state of incomplete pages, removing the burden to manage it from the programmer. In essence we are moving from the "hard way" to the "easy way" in this [breakdown of multi-page forms](https://www.codewithjason.com/rails-multi-step-forms/). When a user finished configuring a `page` the different records would be collected together and submitted to the API as before.
+
+Essentially:
+
+* Refactor our models to be more Rails-standard and switch from ActiveModel (doesn't easily persist locally) to using ActiveRecord (allows us to easily persist locally) on our classes.
+
+* Rather than manually managing data in session for the user, as the user takes an action the data entered via the form is persisted to the PostgreSQL database for forms-admin.
+
+* Use the PostgreSQL database to manage data.
+
+* Drop using cookies for data in session (other than checking if the user is logged in)
+
+## Decision
+
+* We add a new class that is an ActiveRecord model. This is a join table between Forms and Pages in the database.
+* Temporarily, we keep use of `FormObjects` to break up the record into multiple pages and validate at those steps, before saving the record.
+
+## Consequences
+
+### Pros
+
+* removes need for Redis instance
+* it makes the app more standard Rails
+* it removes massive tech debt (where we went from a single-page, zero workflow interface to trying to adapt it to have some workflow)
+* it makes things easier to implement
+* it’s easier to validate
+* data is more stable
+* users don’t lose their data as their workflow diverges
+* as more complexity is added to the workflow, we won’t be adding to that tech debt and working against and increasing brittleness
+* developers don't need to manage incomplete page state "by hand" as we could fallback on active record to save records to the database. This matches how we manage other parts of the form.
+* less likely for data to be deleted accidentally.
+* lets us delete some existing session handling code and tests.
+
+### Cons
+
+* Risk of duplication of business logic between api and admin apps. This is why the concern of the api is to act as the "single source of truth" for forms, and only be concerned with rendering *viewable* forms, both previews and live.


ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,41 @@
...
+
+> Proposed
+
+## Context
+
+Originally forms-admin didn't have a workflow but a single screen for the user to draft their form. This got broken up into a multistage workflow. Data is transferred between screens via cookies but this has limitations, and only worked ok for a linear flow. Now that we are approaching diverging workflows, we can't support this any longer. We are also currently only using ActiveModel for our classes.
+
+We should:
+
+* Refactor our models to be more Rails-standard and switch from ActiveModel only to ActiveRecord.
```
lfdebrux, line 15: I'm a bit new to Rails, so it's not obvious to me what this means; is there an advantage of switching from ActiveModel only to ActiveRecord other than being more Rails-standard?

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,41 @@
...
+
+> what the team has decided to do
+
+## Consequences
+
+### Pros
+
+* removes need for Redis instance
+* it makes the app more standard Rails
+* it removes massive tech debt (where we went from a single-page, zero workflow interface to trying to adapt it to have some workflow)
```
lfdebrux, line 33: I like removing tech debt!

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Context
+
+Originally forms-admin didn't have a workflow but a single screen for the user to draft their form. This got broken up into a multistage workflow. Data is transferred between screens via cookies but this has limitations, and only worked ok for a linear flow. Now that we are approaching diverging workflows, we can't support this any longer. We are also currently only using ActiveModel for our classes.
+
+We should:
+
+* Refactor our models to be more Rails-standard and switch from ActiveModel only to ActiveRecord.
+
+* Use the PostgreSQL database to manage data.
```
lfdebrux, line 17: Is this a standard Rails thing? Is there documentation on how to do this we can refer to?

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,41 @@
...
+* removes need for Redis instance
+* it makes the app more standard Rails
+* it removes massive tech debt (where we went from a single-page, zero workflow interface to trying to adapt it to have some workflow)
+* it makes things easier to implement
+* it’s easier to test
+* data is more stable
+* users don’t lose their data as their workflow diverges
+* as more complexity is added to the workflow, we won’t be adding to that tech debt and working against and increasing brittleness
+
+### Cons
```
lfdebrux, line 40: Are there any cons at all? 😁 Do we need to do garbage-collection in PostgreSQL? Do we need to put other work on hold while doing this refactor? What about deploying, does that need some thought?

lfdebrux (commented):
Saw this in my GitHub notifications and couldn't resist a peek, I hope you don't mind me leaving some early comments 😅  I appreciate it's still a work in progress so feel free to ignore them.

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
chao-xian: Good point! I'll expand on that.

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
chao-xian: This is why I'm drafting the ADR! I can't think of any... The main one might be that we start to duplicate some of forms-api locally, however I feel that's a cost worth taking and we can keep api as the more "source of truth" of forms for rendering and published state. It's possible it'll make it easier to merge the two, should we decide to do.

aliuk2012: I think one cons might be that we are "splintering" the data into much smaller chunks which would then have to be "pieced" together before sending data to forms-api.

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
chao-xian: We can either store sessions in db, or migrate from that (later!) and only have session for logged in user and move to a more author per form version model

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,47 @@
...
+
+We should:
+
+* Refactor our models to be more Rails-standard and switch from ActiveModel (doesn't easily persist locally) to using ActiveRecord (allows us to easily persist locally) on our classes.
+
+* 
+
+* Rather than manually managing data in session for the user, as the user takes an action the data entered via the form is persisted to the PostgreSQL database for forms-admin.
+
+* Use the PostgreSQL database to manage data.
```
thomasiles, line 23: My understanding of the ADR is this:

Users configure `pages` in forms admin over multiple steps (sometimes called a wizard 🧙!). `Pages` are only saved in a form when a user has completed all the required steps. Different answer types have difference steps. 

For example, the user chooses the "date" answer type, then on the following screen selects that it will be a date of birth and on the last screen enters the question text. The user then saves the `page`, at which point the `page` is added to the `form`.

Currently we use the [form object pattern](https://dev.to/drbragg/rails-design-patterns-form-object-4d47#:~:text=Form%20Object%20Pattern,consistent%20API%20in%20our%20code.) to manage the intermediate state of pages as they are being created.

 This gives each screen a [model](https://github.com/alphagov/forms-admin/tree/main/app/forms/pages), [controller](https://github.com/alphagov/forms-admin/tree/main/app/controllers/pages) and [view](https://github.com/alphagov/forms-admin/tree/main/app/views/pages).

We store the state of incomplete `pages` in the session hash. The code updating this hash is spread across the form objects and controllers. When adding or modifying steps in the `page` journey, it's up to the programmer to make sure the session is updated correctly, such as making sure it's cleared in the right places.

As we've added more features for `pages` such as extra input types, conditions and additional guidance, managing state in the session hash has become harder. We've had issues around producing inconsistent data, using a mix of strings and symbols and deleting data at the wrong point.

Also, as the session hash is currently stored in user's cookie, there is a size limit of 4kb. This is not an immediate issue but could pose difficulties when a `page` has many question options, long guidance text or we start using the session in other parts of the application.

This ADR proposes that we stop using the forms object pattern to create pages. Instead each step is represented using an activeRecord model and database table. This means we no longer need to use the session hash to maintain the state of incomplete pages, removing the burden to manage it from the programmer. In essence we are moving from the "hard way" to the "easy way" in this [breakdown of multi-page]forms(https://www.codewithjason.com/rails-multi-step-forms/). When a user finished configuring a `page` the different records would be collected together and submitted to the API as before.

I think the pros would be:
- we don't need extra storage for the session to solve the 4kb issue as the session would no longer be storing any significant state.
- developers don't need to manage incomplete `page` state "by hand" as we could fallback on active record to save records to the database. This matches how we manage other parts of the form. This should make it easier to  validate and less likely to be deleted accidentally. This should also let us delete some existing session handling code and tests. 

I think the challenges are:
- we need to think about how the lifetime and scope of incomplete `pages`. Currently, using the session, means that pages are tied to a user's browser and are removed when the session is deleted. Storing them in the admin database means we need to know when it's safe to delete them and how to handle them across multiple users, or a single user with multiple tabs open.
- given the above we need to weigh the pros/cons of storing this data in the admin at all. If the API is already storing `forms` and `pages`, should it store incomplete `pages` too?
- Is this any better than being more disciplined about storing the incomplete pages state within the session. Currently we just access the session as hash. Would it be better use objects to store the state within the hash?

I see [Al's spike](https://github.com/alphagov/forms-admin/pull/631) as being parallel to this ADR. The spike uses postgres as a session store. This is no different than using redis as a session store expect that we don't need infra to do anything for us 😄 This is obviously still a win over storing large bits of state in cookeis and cleaning up the code we use to access the session still seems like a great step towards this ADR. I think linking to the spike underneath the top paragraphs might confuse things a bit though.

Does the above make sense? Have I got the right end of the stick about what the ADR is suggesting?

chao-xian: Thanks for writing down what I'd been trying to convey but in a much better way @thomasiles! And for sharing those links. If you don't mind, I'm going to incorporate that into the ADR itself? 

Just as an initial response right at the end of the day:

Al's spike is very early stuff and the first phase of doing what the ADR ultimately wants to achieve. 

I'd like to move away from thinking in terms of constructing per "session" and more towards an author is drafting a thing and their session is just for being logged in kind of paradigm. The act of "publishing" to the api is the point at which we delete from the database associated data if we need to, is how I envisage it.

The API in my view should only store what's going to get rendered to the user. It makes for a clearer separation where the admin app is for drafting, the api is for storing published forms, and the runner for rendering those -  but that's not a concern for now.

Until then, shifting the session into the database is a preliminary step and gives us some safety as we can then prune stuff out of being "in session" as we port the models to ActiveRecord and stored in the database. We think detailed guide is a good candidate.

Hope that makes sense...

thomasiles: yeah I think that makes sense.

I'm tempted to say we don't need an ADR for changing session storage and patterns internal to forms-admin. I think it's something useful to discuss and work through as a group because there is quite a lot to it. But I don't think it means any major architecture changes on its own.

I think the concept of using the API only for "published forms" (we call them "made live" forms in the forms team) would make sense as an ADR on its own. It's [option 2 of ADR004](https://github.com/alphagov/forms/blob/main/ADR/ADR004-store-form-schemas-in-an-api.md) and if we are making a decision, it would be useful to reference it.

I think it would make writing an ADR easier and it would also allow James to make more of a contribution. There is a danger that it sounds like we are discussing a detail of Rails programming style rather than application architecture.

It's something we have discussed (at length!) before but I'm still in favour of going over it again.

ADR/ADR022-use-postgresql-for-form-creation-in-form-admin.md
```diff
@@ -0,0 +1,70 @@
...
+## Strategy
+
+* We add a `FormPages` class that is an ActiveRecord model. This is the join table between Forms and Pages in the database.
+* Temporarily, we keep use of `FormObjects` to break up the record into multiple pages and validate at those steps, before saving the record.
+
+## Decision
+
+> Accepted?
+
+## Consequences
```
DavidBiddle, line 52: If we hypothetically wanted to monolith-ise our apps in future, do we think this change would make that easier, harder, or make no difference?

(my instinct is that this might make it slightly easier but you'll have a clearer idea than I do)

chao-xian: It probably will be easier _if_ we decided to Monolithi-cise (but from pov I don't want to).

danworth (approved):
This has been well considered by those in the team in a position to do so and its currently half-way to be being implemented. So long as nothing has come up during that partial implementation that might make us reconsider, then lets proceed ⏩ 🏅

PR #113, opened 2023-11-01
Update version-1.md - minor style tweaks

C-Harry:
Have made a few style tweaks and amended a couple of typos

# How to review

Check you agree with the changes.

### DIFF
--- a/design/features/form-metrics/version-1.md
+++ b/design/features/form-metrics/version-1.md
@@ -29,20 +29,20 @@ ___
 
 #### What have we been sending  
 
-- In order of importance for service teams 
+- In order of importance for service teams: 
     1. Submissions
     2. Completion rate
     3. Question funnel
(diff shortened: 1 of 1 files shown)


design/features/form-metrics/version-1.md
```diff
@@ -29,100 +29,100 @@ ___
 
 #### What have we been sending
```
C-Harry, line 30: I'm a bit confused about this section as not all these bullet points refer to things we've been sending, do they? (E.g. ability to download raw data and 'Some of the Google Analytics metrics such as device type [...])

christophercameron-ixd: Yeah I think this was the things Anne wanted/expected us to provide once in place - copying mural comments over was a bit difficult with this as they were muddled. Will move them to a different section. Maybe something like future expectations

PR #114, opened 2023-11-06
Document early access journey designs

hazalarpalikli:
# PR Checklist

---

# What

# How to review

For example:

# Who can review

### DIFF
--- a/design/features/early access journey
+++ b/design/features/early access journey
@@ -1,0 +1,49 @@
+# Editing pages when looking at a preview of a form
+
+## Status
+
+Date created: *2023-11-06*  
+
+Trello card:https://trello.com/c/I8HETm57/1041-design-the-early-access-upgrade-mou-journey-ready-for-user-research 
+Mural working board: https://app.mural.co/t/gaap0347/m/gaap0347/1691659258399/af093136c3f86be05860120cb92d6810e98f3643?sender=u61ceb93b0f861033ca7c7448 
+
+___
+
+## Contents
+
+- [Early access journey](#early-access-journey)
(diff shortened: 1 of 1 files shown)


design/features/early access journey
```diff
@@ -0,0 +1,49 @@
+# Editing pages when looking at a preview of a form
```
hannahkc, line 1: Think this is the wrong heading?

design/features/early access journey
```diff
@@ -0,0 +1,49 @@
...
+The changes are in the product. 
+Prototype can be found here: https://forms-prototypes-pr-201.herokuapp.com/product-pages/ 
+
+<br>
+
+___
+
+<br>
+
+[Back to the top](#editing-pages-when-looking-at-a-preview-of-a-form)
```
hannahkc, line 49: think this is the wrong link?

design/features/early access journey
hazalarpalikli: Yeah i need to delete this

design/features/early access journey
hazalarpalikli: It is the right link, i have added the protoype to the forms protoype.

hannahkc: meant this: 

[Back to the top](#editing-pages-when-looking-at-a-preview-of-a-form)

which won't work when you correct the heading.

hannahkc: Also I think the idea with this is to record designs here so we can look back at previous versions and why they were the way they were. The prototype is only temporary and can be replaced, deleted, moved. We should only be using the prototype for testing things really.

PR #118, opened 2023-11-13
Update auth diagram with session timeout

thomasiles:
Trello: https://trello.com/c/PET8LyHH/1120-update-authentication-sequence-diagrams

I've made a few changes to this diagram to try and make it clearer what's happening. I'm happy to revert/change any of them if it's not helping.

In this PR, I've:

- Added the missing redirect to forms admin/auth/auth0.

- Split Auth0 into two actors to reflect the different roles it's playing in our system.
Originally this was to show the difference between our login session and the [auth0 session](https://auth0.com/docs/manage-users/sessions/session-layers).

  [It turns out this doesn't need to be documented
yet](https://trello.com/c/rmk3P0kO/1299-fix-auth0-single-sign-on), but I think it still might be useful. It would be good to compare with the google workspace diagram.

  It could be that auth0 is outside our control so we should treat it as a black box.

- Changed the references to session to be more precise about what's happening. Talking in terms of the cookie values is lower level but gives [this ticket](https://trello.com/c/yi3v9Mgo/1121-require-forms-admin-users-to-re-authenticate-after-24-hours) a clearer description.

  Session is the right term but it is also implies more is happening behind the scenes, which I don't think there is here.

### DIFF
--- a/diagrams/sequence-diagrams/authentication.md
+++ b/diagrams/sequence-diagrams/authentication.md
@@ -8,44 +8,53 @@ sequenceDiagram
     participant browser as web browser
     participant forms as GOV.UK Forms
     participant auth0 as Auth0
+    participant auth0-ui as Auth0 Universal Login
     participant ses as Amazon SES
 
     user ->> browser: visit GOV.UK Forms
     browser ->> forms: HTTP GET
-    forms ->> forms: no session cookie found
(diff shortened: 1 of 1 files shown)


james-sheppard-gov-uk (approved):
Thanks for making those changes, looks good @thomasiles !

PR #126, opened 2024-01-08
ADR for replacing form-id in urls

SamJamCul:
Rendered view https://github.com/alphagov/forms/blob/adr-reaplacing-form-ids-in-urls/ADR/ADR023-use-uid-in-urls.md

This ADR covers the changes that [this spike](https://trello.com/c/BruuEN5Z/1225-spike-change-ids-in-urls-of-forms-to-not-be-the-primary-key) has led to. 

The idea is we don't want to expose our database IDs in our URLs, and in order to do this we need to use something other than our IDs to retrieve our forms from the API.

### DIFF
--- a/ADR/ADR023-use-uid-in-urls.md
+++ b/ADR/ADR023-use-uid-in-urls.md
@@ -1,0 +1,37 @@
+# ADR23: Use UIDs in URLs to replace form-id
+
+Date: 08/01/24
+
+## Status
+Accepted
+
+## Context
+
+See this spike for details about the required changes https://trello.com/c/BruuEN5Z/1225-spike-change-ids-in-urls-of-forms-to-not-be-the-primary-key
+
+We need to switch away from putting database ids in our URLs. Currently, form urls look something like this:
+`https://submit.forms.service.gov.uk/form/1234/apply-for-an-apple/5678`
+
+Where `1234` is the form-id.
+
+We instead want to change our URLs so that they look like this: 
+
+`https://submit.forms.service.gov.uk/form/51f5b6f7/apply-for-an-apple/5678`
+
+where the 8 char hex code `51f5b6f7` has replaced the form-id.
+
+We also have to handle existing form URLs, and how we want to maintain them.
+
+In order to achieve this change, we'll add an `external_id` to all of our forms, which will be used to identify from the exterior of the API. For pre-existing live forms, they will be given their current `id` as their `external_id` which will maintain their current URLs as a result. 
+
+For example, a new form will have it's `external_id` as `51f5b6f7` whereas an old form will use it's current `id` of `1234` as its `external_id`. 
+
+## Decision
+
+We will be using an external ID to retrieve forms from forms-api
+
+## Consequences
+
+URLs will no longer expose our database primary key IDs, and our API will not find objects by their primary key ID. Instead, their external ID will need to be used. 
+
+URLs for forms-runner and forms-admin will use the external ID, which will change their appearance. 


ADR/ADR023-use-uid-in-urls.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Redirecting from old URLs
+
+We will maintain a list of old URLs mapped to new URLs for live forms that are in use. This will be limited to whatever the current set of live forms is (something like 22 live forms at the moment), and it won't grow. 
+ 
+We'll monitor these redirects to see their usage, and we can phase them out if we see that no one uses them after a certain point. 
+
+The redirect mapping will be 1-to-1, so anyone visiting `form/{current-form-id}/apply-for-an-apple/5678` will redirect to `form/{unique-id}/apply-for-an-apple/5678`. Even if we migrate our data so that forms are given new IDs, the old ID mapping will remain so as not to break bookmarked links.
+
+### Existing forms retain their current URLs
```
chao-xian, line 38: We discussed this as being sub-optimal in a sense as this will confuse future devs/users.

SamJamCul: Yep, I'll remove it in favour of the redirect system

ADR/ADR023-use-uid-in-urls.md
```diff
@@ -0,0 +1,43 @@
...
+
+`https://submit.forms.service.gov.uk/form/51f5b6f7/apply-for-an-apple/5678`
+
+where the 8 char hex code `51f5b6f7` has replaced the form-id, or like this:
+
+`https://submit.forms.service.gov.uk/form/03998919-6bd1-48f3-a788-b3e3d5c9f822/apply-for-an-apple/5678`
+
+where the uuid `03998919-6bd1-48f3-a788-b3e3d5c9f822` has reaplced the form-id.
+
+We also have to handle existing form URLs, and how we want to maintain them. We will give all forms old and new a new random character URL, and redirect from all existing form's old URLs.
```
lfdebrux, line 27: 👍

ADR/ADR023-use-uid-in-urls.md
```diff
@@ -0,0 +1,43 @@
...
+We need to switch away from putting database ids in our URLs. Currently, form urls look something like this:
+`https://submit.forms.service.gov.uk/form/1234/apply-for-an-apple/5678`
+
+Where `1234` is the form-id.
+
+We instead want to change our URLs so that they look either like this: 
+
+`https://submit.forms.service.gov.uk/form/51f5b6f7/apply-for-an-apple/5678`
+
+where the 8 char hex code `51f5b6f7` has replaced the form-id, or like this:
```
lfdebrux, line 21: How many different possibilities does an 8 char hex code give us?

DavidBiddle: I think it would be 16^8 = 4,294,967,296

DavidBiddle: Which for scale is about half the total number of Notify messages has ever sent

lfdebrux: Nice!

ADR/ADR023-use-uid-in-urls.md
chao-xian: Ah hang on I've had a chat with Dan actually and I get the idea of it being the minimal change for now and we can convert the IDs in future if we need. Sorry I've been in meetings all day and not had a chance to look at this. Worth reinstating if you still have that change?

ADR/ADR023-use-uid-in-urls.md
chao-xian: Yes let's go with the 8-char hex as that's also easier to type

SamJamCul: Al and I were spiking it out, we can use `SecureRandom.uuid[0, 8]` where we're using the first 8 chars of a UUID. We still

lfdebrux: I think there is a small risk of generating a rude word, if we're worried about that we could look at something like [sqids](https://sqids.org/ruby) as suggested by @AP-Hunt.

lfdebrux: I had a thought earlier, how many forms do we generate in a year with our end to end tests as they currently work? 🤔 They currently run every time we do a release of forms-admin, forms-api or forms-runner, and create (and delete) one new form on each run.

If we don't care about collisions between different stages then that's equal to the number of PRs merged in a year for each of those repos. Since 2023-01-10, we've merged 569 PRs to forms-admin, 211 for forms-api, and 327 to forms-runner. That's 1107 releases a year. Sanity checking that, there are currently 758 forms in the production database, and at time of writing the highest id in the database is 1873, so it seems reasonable that the end to end tests have created over a 1000 forms since they started running. 

For niceness, let's say we're currently creating 1024 forms a year for our end to end tests. So assuming things continue in the same way, and we only care about production, I _think_ we're looking at consuming (1024/16^8)*100 percent , or 0.000002 % of the possibility space... so I think we're good 😁

ADR/ADR023-use-uid-in-urls.md
```diff
@@ -0,0 +1,37 @@
...
+## Context
+
+See this spike for details about the required changes https://trello.com/c/BruuEN5Z/1225-spike-change-ids-in-urls-of-forms-to-not-be-the-primary-key
+
+We need to switch away from putting database ids in our URLs. Currently, form urls look something like this:
+`https://submit.forms.service.gov.uk/form/1234/apply-for-an-apple/5678`
+
+Where `1234` is the form-id.
+
+We instead want to change our URLs so that they look either like this:
```
DavidBiddle, line 17: Don't need this now that we've removed the other option:

```suggestion
We instead want to change our URLs so that they look like this: 
```

aliuk2012:
I think it might be worth adding mentions here what we will do with existing draft/live forms, we will be using the old original ID for this forms and only new ones will use the new 8digit hex

SamJamCul:
> I think it might be worth adding mentions here what we will do with existing draft/live forms, we will be using the old original ID for this forms and only new ones will use the new 8digit hex

I've added a bit that covers this:

> In order to achieve this change, we'll add an `external_id` to all of our forms, which will be used to identify from the exterior of the API. For pre-existing live forms, they will be given their current `id` as their `external_id` which will maintain their current URLs as a result.

ADR/ADR023-use-uid-in-urls.md
SamJamCul: good spot!

DavidBiddle (approved):
Looks good! ADR makes sense and accurately records the things we've discussed in tech catchups

ADR/ADR023-use-uid-in-urls.md
```diff
@@ -0,0 +1,37 @@
...
+
+For example, a new form will have it's `external_id` as `51f5b6f7` whereas an old form will use it's current `id` of `1234` as its `external_id`. 
+
+## Decision
+
+We will be using an external ID to retrieve forms from forms-api
+
+## Consequences
+
+URLs will no longer expose our database IDs, and our API will not find objects by their ID. Instead, their external ID will need to be used.
```
chao-xian, line 35:
```suggestion
URLs will no longer expose our database primary key IDs, and our API will not find objects by their ID. Instead, their external ID will need to be used. 
```

SamJamCul: Good one, I've fixed it up

danworth (approved):

chao-xian (approved):

PR #128, opened 2024-01-15
Add ADR Store allowed email domain list as file in forms deploy

lfdebrux:
# What

Trello ticket: https://trello.com/c/7Wfx1tI5/1255-spike-how-to-support-allow-deny-lists-from-a-shared-source

# How to review

For example:

# Who can review

### DIFF
--- a/ADR/ADR024-store-allowed-domain-list-as-file-in-forms-deploy.md
+++ b/ADR/ADR024-store-allowed-domain-list-as-file-in-forms-deploy.md
@@ -1,0 +1,134 @@
+# ADR024: Store the list of approved email address domains as a configuration file in the forms-deploy repo
+
+Date: 2024-01-29
+
+## Status
+
+Accepted
+
+## Context
+
+We want to allow anyone with a government email address to be able to use GOV.UK Forms to create forms. Most, email addresses in government have a domain that ends with `.gov.uk`, but not all. For this reason we need a list of allowed email address domains.
+
+However, not all government organisations want to allow anyone in their organisation to use GOV.UK Forms. For this reason we need a list of email address domains that have access restrictions.
+
+Currently in our source code there are two different lists of email address domains in two different repos, with some duplication:
+
+- [forms-deploy/infra/deployments/production/auth0/main.tf](https://github.com/alphagov/forms-deploy/blob/879f20c24a9dc1ed5d7037239a1f22c335d84820/infra/deployments/production/auth0/main.tf)
+- [forms-admin/app/models/user.rb](https://github.com/alphagov/forms-admin/blob/0028969189cf69682c54fb9422ca129012fbf3ee/app/models/user.rb) includes a list of restricted domains
+
+Additionally [forms-admin/app/validators/allowed_email_domain_validator.rb](https://github.com/alphagov/forms-admin/blame/0028969189cf69682c54fb9422ca129012fbf3ee/app/validators/allowed_email_domain_validator.rb) defines a validator that is supposed to check whether an email address has a domain in the list of allowed domains; however it does this currently by assuming that the user's email address is in the list of allowed domains. This was done to avoid further duplication, but limits what users can do.
+
+We would like a different approach to storing this data, one that allows us to have a single source of truth that is easy to maintain.
+
+We may also in future want to be able to use the association between an organisation and it's email address domains to be able to provide more granular access control, and/or stricter validation, and/or protective monitoring. We currently have data about government organisations in the databases for our service, and an existing model for organisations in our code.
+
+Ideally, we would have a email address domain model. Each domain would belong to an organisation, and each organisation would have have zero or more domains. The organisation model would have the option of whether new users in that organisation would have access to GOV.UK Forms. The lists of allowed and restricted email address domains would be derived from the access control option on associated organisations.
+
+```mermaid
+classDiagram
+    class Organisation {
+        boolean new_users_have_access
+    }
+
+    class EmailAddressDomain {
+        .allowed_email_address_domains()
+        .restricted_email_address_domains()
+    }
+
+    Organisation "0..1" --> "*" EmailAddressDomain
+
+    class User {
+        string email_address
+        boolean has_access
+    }
+
+    Organisation "0..1" --> "*" User
+```
+
+The obvious option for managing the email address domain data would be to store it in a database alongside the organisation data, as part of the service, and manage it like any other database records in a Rails app, with Rake tasks, a RESTful API, and/or web forms.
+
+However, there are two downsides to this approach: One, this is configuration data that is defines how the service is run, and should be stored in a way which allows control, auditing and replication independent of the runtime data. Two, these lists need to be used in two different places, in the identity provider that authenticates users email addresses as well as the `forms-admin` app.
+
+```mermaid
+%%{init: {"flowchart": {"defaultRenderer": "elk"}} }%%
+flowchart TD
+    subgraph forms-admin
+    start([User visits GOV.UK Forms])
+    --> Auth0
+
+    subgraph Auth0
+    login_id_prompt[Ask user for email address]
+    -->
+
+    restrict_users_to_allowed_domains{Is the email address domain<br>in the allow list?}
+    -- No -->
+    unauthorised_email_domain([Show error<br>'Enter a government email address'])
+
+    restrict_users_to_allowed_domains -- Yes -->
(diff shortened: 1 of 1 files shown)


ADR/ADRYYY-store-allowed-domain-list-as-file-in-forms-deploy.md
```diff
@@ -0,0 +1,126 @@
...
+The identity provider we are using currently is Auth0; it is configured using Terraform code in the `forms-deploy` repository. The `forms-admin` app is built and run as a container. The container image is built from code in the `forms-admin` repository and deployed as part of a release pipeline. This pipeline is defined using code in the `forms-deploy` repo, and also itself retrieves the `forms-deploy` repo to determine how the `forms-admin` app should be deployed. Getting data from the databases into Auth0 is more difficult compared to getting data into the app during deployments. This suggests a solution where a single source-of-truth ouside of the database is access by code in the `forms-deploy` repo both when Auth0 is configured by terraform and when the deployment pipeline is run.
+
+## Decision
+
+> what the team has decided to do
+
+Add configuration file(s) (YAML or JSON) to forms-deploy repo with list of domains and associated organisations
+    - Suggest folder `forms-deploy/config/data`
+
+Add Rake task to load domain configuration into database from file or environment variable
```
DavidBiddle, line 112: Could we run this rake task on a regular schedule (and is that something we'd want)?

lfdebrux: My thought was that the Rake task would run every time forms-admin is deployed, that feels like the correct behaviour to me, what do you think? I'll try and update this doc to make that clearer if you agree.

DavidBiddle: I think line 103 covers this sufficiently now that I've read it properly! Was mainly interested in whether this was an automated task or one to be run manually whenever we update the list, so my question is answered :+1:

jamie-o-wilkinson, line 112: Initially I thought storing the email domains in the database would be overcomplicating this versus just using parameter store, at least until we have a concrete need for something more sophisticated.

However, that need for something more sophisticated is rapidly approaching if we are going to be automating some part of the organisation assignment based on email domain. If we need that functionality, I think what is proposed here is going to work well.

ADR/ADRYYY-store-allowed-domain-list-as-file-in-forms-deploy.md
```diff
@@ -0,0 +1,126 @@
...
+    - See spike in forms-admin branch https://github.com/alphagov/forms-admin/compare/main...ldeb-spike-domains-read-into-database
+
+Change the pre-deploy configuration to inject the configuration for the environment from forms-deploy repo and then run the load domain configuration task
+    - See `forms-deploy/infra/modules/shared/scripts/db-migrations-pre-deploy.sh`, `forms-deploy/infra/modules/shared/scripts/db-migrations-ecs-run-task-input.jq`
+
+## Consequences
+
+> both positive and negative consequences of the decision
```
jamie-o-wilkinson, line 120: Should we mention something here about the risk that the domains table might be updated without the corresponding change in the config? i.e. we'd have to ensure that the config remains the single source of truth.

ADR/ADRYYY-store-allowed-domain-list-as-file-in-forms-deploy.md
```diff
@@ -0,0 +1,126 @@
...
+        string email_address
+        boolean has_access
+    }
+
+    Organisation "0..1" --> "*" User
+```
+
+The obvious option for managing the email address domain data would be to store it in a database alongside the organisation data, as part of the service, and manage it like any other database records in a Rails app, with Rake tasks, a RESTful API, and/or web forms.
+
+However, there are two downsides to this approach: One, this is configuration data that is defines how the service is run, and should be stored in a way which allows control, auditing and replication independent of the runtime data. Two, these lists need to be used in two different places, in the identity provider that authenticates users email addresses as well as the `forms-admin` app.
```
danworth, line 55: I think the challenge around this arises because Auth0 is not only performing authentication (does the user have control of the email address they claim and therefore we're confident they are who they say they are) but also some initial authorisation (is that email domain allowed as a user of forms). If we separated those concerns so that Auth0 only performed the authentication and forms-admin only performed authorisation then it is only forms-admin that needs the allowed list or knowledge of email addresses and their authorisation restrictions and it's a simpler problem to solve (there is only one source used by one consumer). 

How feasible would such an approach be with how Auth0 is setup?

Your diagram might then look something like:
```mermaid
%%{init: {"flowchart": {"defaultRenderer": "elk"}} }%%
flowchart TD
    subgraph forms-admin
    start([User visits GOV.UK Forms]) -->
    login_id_prompt[Sign in page, ask user for email address] -->
    restrict_users_to_allowed_domains{Is the email address domain<br>in the allow list?}
    -- No -->
    unauthorised_email_domain([Show error<br>'Enter a government email address'])
    restrict_users_to_allowed_domains -- Yes -->Auth0
    subgraph Auth0
    
    authenticate_email_address[[Check that the user has<br>access to that email inbox]]
    end
    authenticate_email_address -->
    find_or_create_user{Has the user signed in before?}
    find_or_create_user -- No -->
    create_user[Create a new user record]
    -->
    deny_users_with_restricted_domains{Is the email address domain<br>in the restricted list?}
    deny_users_with_restricted_domains -- No -->
    set_has_access_true[Allow access for the new user]
    --> has_access
    deny_users_with_restricted_domains -- Yes -->
    set_has_access_false[Deny access for the new user]
    --> has_access
    find_or_create_user -- Yes -->
    has_access{Does the user have access?}
    has_access -- No -->
    access_denied([Show error 'You do not have permission'])
    has_access -- Yes -->
    govuk_forms_home_page([Welcome to GOV.UK Forms])
    end
```

lfdebrux: @danworth yes, not having Auth0 need to know the allowed email domains would simplify things in terms of where we can store the list of domains.

However, if we want to ask the user their email address before sending them to Auth0 I think it would complicate the Auth0/identity provider setup... 😅 

There are also some benefits to keeping the email domain configuration in forms-deploy anyway; @AP-Hunt made the case to me that it's configuration data, not something that should live (solely) in the database or in the source code. If we agree that we want to keep the email domain list in forms-deploy then it's not much additional complexity to share the list with Auth0. We might also at some point want to revisit restricting the email addresses SNS can send to [[1](https://github.com/alphagov/forms-deploy/pull/450)], in which case having the list in forms-deploy is again necessary even if we're not restricting in Auth0.

ADR/ADRYYY-store-allowed-domain-list-as-file-in-forms-deploy.md
lfdebrux: I've added something about this now.

ADR/ADRYYY-store-allowed-domain-list-as-file-in-forms-deploy.md
danworth: Yeah I appreciate that its already gone in this direction and it may not be pragmatic at this point in time to reverse it so I'll resolve this; we should just consider very carefully before putting any more authorisation into the authentication provider.

ADR/ADR024-store-allowed-domain-list-as-file-in-forms-deploy.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Configuration of domains will be version controlled.
+
+Negative consequences:
+
+- There could be a mismatch between data in the app database and the configuration file.
+
+- Developers will need to be involved in updating the domain list.
+
+A change to the domain list will require an re-deploy.
```
DavidBiddle, line 134:
```suggestion
A change to the domain list will require a re-deploy.
```

DavidBiddle (dismissed):
Really well written and explained ADR. The decision seems sensible and matches what we've discussed on the tech team previously.

Spotted one tiny typo but otherwise looks great 🎉

ADR/ADR024-store-allowed-domain-list-as-file-in-forms-deploy.md
lfdebrux: Nice spot, thanks!

DavidBiddle (approved):

PR #129, opened 2024-01-16
ADR for paramaterising terraform roots

alice-carr:
Trello Ticket: https://trello.com/c/QC5D7hgL/317-write-an-adr-for-terraform-restructuring

# PR Checklist


ADR/ADR023-paramaterise-common-terraform-roots.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Consequences
+
+### Pros
+* creating a new environment with this terraform structure will be quicker and easier, since there is less configuration that will need to be done
+* we have removed repeated code, making the terraform easier to read
+* this reduces the chance that environments will differ due to human error, since paramaterised inputs are stored together within the root
+
+### Cons
+* manually deploying terraform is now a bit more complicated. We have remedied this with the addition of a [makefile](https://github.com/alphagov/forms-deploy/blob/main/Makefile) which can now be used to run terraform locally
```
danworth, line 88: I agree with all the pros and this is a nicely written ADR. I mostly agree with this con but I think its currently underplaying it a bit. I wouldn't say that the additional complexity has been "remedied" as such. The make file and associated bash script tooling makes it _easier_ to perform the terraform operations by encapsulating the additional complexity, but it still remains and the tooling will need to be understood, maintained and modified as necessary going forward.

alice-carr: thanks Dan - I've updated this

cadmiumcat (approved):
I think this reflects well what we have done

ADR/ADR023-paramaterise-common-terraform-roots.md
```diff
@@ -0,0 +1,88 @@
+# ADR023: Paramaterise Common Terraform Roots
```
aliuk2012, line 1: ADR23 has already been taken along with ADR24 which is in another PR that Laurence has created

AP-Hunt (dismissed):
Good job capturing the cons

ADR/ADR023-paramaterise-common-terraform-roots.md
alice-carr: thanks Al, I've updated the number

AP-Hunt (approved):

PR #130, opened 2024-01-17
2024 01 17 new features readme layout and content

christophercameron-ixd:
# PR Checklist

- [X] Set yourself as the Assignee
- [X] Tag anyone you would like to review, or @forms-design or @forms-devs
- [X] Fill in the template below

---

# What

* added a new `README.md` to the `/design` directory to add a little explanation of what it is for and how we use it
* updated the `/features/README.md` file to remove mention of not testing them in the prototype (because we have for some), and added links and descriptions for all the features we have documented
* changed the `prototype-tested.md` file names for the prototype version to `README.md` so the content displays if the user goes to a version directory
* removed a random .DS... file
* removed unused feature-template and prototype-version templates, we can just reuse existing directories and files to start from

# How to review

3. Have I missed any other unnecessary files? 

# Who can review  

Anyone in the team can review, especially as most of it is file renames and removals.   
But content review of the new `design/README.md` and updated content of the `features/README.md` would be welcomed!


C-Harry (approved):
Looking good I think!

PR #133, opened 2024-01-29
Add link to drafted content for question previews

hannahkc:
# What
I've added a link to a document of content for previewing a question for each different answer type. This is just a temporary measure for now - until the content is updated in the prototype when we can document it properly.

# How to review
1. check the link works. 
2. check Hannah hasn't made any typos or errors. 
3. tell Hannah if you think this is a terrible thing to do. 

# Who can review
Anyone. Only needs one review.


C-Harry (approved):
Looks fine to me.

PR #140, opened 2024-02-22
Add ADR For Scheduled Smoke Tests

danworth:
ADR for how we'll run scheduled smoke tests.

Please see the Trello Ticket for the spike into running scheduled smoke tests for more information including working infra and capybara code: https://trello.com/c/oRlozBpd/1393-spike-smoke-tests

### DIFF
--- a/ADR/ADR027-scheduled-smoke-tests.md
+++ b/ADR/ADR027-scheduled-smoke-tests.md
@@ -1,0 +1,43 @@
+# ADR027: Scheduled Smoke Tests
+
+Date: 2024-02-22
+
+## Status
+
+Accepted
+
+## Context
+
+We want to run scheduled smoke tests that check critical user journeys like a
+user completing and submitting a form. If a smoke test fails then the team
+should be alerted through various channels depending on the required response
+(e.g. Slack or Pager Duty out of hours).
+
+## Decision
+
+1. Use Ruby and Capybara to write the test code to maintain consistency with
+   the end-to-end and application code already in use.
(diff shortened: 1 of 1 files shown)


ADR/ADR027-scheduled-smoke-tests.md
```diff
@@ -0,0 +1,43 @@
...
+should be alerted through various channels depending on the required response
+(e.g. Slack or Pager Duty out of hours).
+
+## Decision
+
+1. Use Ruby and Capybara to write the test code to maintain consistency with
+   the end-to-end and application code already in use.
+2. Use AWS CodeBuild to run the tests from an image containing the smoke test
+   code along with Chrome and ChromeDriver; this is consistent with how the
+   end-to-end tests are run in our pipelines during deployment. AWS Synthetic
```
chao-xian, line 22: Is it worth mentioning AWS Synthetic Canaries here? Also I'm not personally keen on hooking into another bespoke AWS thing which would potentially make it less portable. Being able to use any other build tool to execute the smoke tests for us feels a better position.

chao-xian (approved):
Looks good! Just one thought. Thanks for writing this!

ADR/ADR027-scheduled-smoke-tests.md
danworth: I've mentioned Canaries there because someone in future might go "what about Canaries?" and then they'll see that when we made this decision Canaries did not support Ruby so it wasn't an option.

PR #146, opened 2024-03-13
2024 02 21 prototype version 4 updates

christophercameron-ixd:
# PR Checklist

- [X] Set yourself as the Assignee
- [X] Tag anyone you would like to review, or @forms-design or @forms-devs
- [X] Fill in the template below

---

# What

This update relates to trelllo card https://trello.com/c/P3Lr5u8y/1396-documentation-fixes-update-prototype-version-4

- Updated what we found (which didn’t really seem to be a huge amount)
- Removed unnecessary formatting and content
- Added link to the UR GitHub (this won’t work until merged as the filename was changed after I branched this version) 
- Added a bit more of a detailed change write up for the “edit question” screen and the minimal feedback we got from the round of testing done on it 

# How to review

3. Is there anything you think I’ve missed? 
4. Were there opportunities or decisions we made off the back of this version? 

# Who can review

Anyone can review, but it’d be great if @hannahkc can take a look I might have miss remembered/missed something we talked about.


oliverquinlan (approved):
All looks good to me!

design/prototype-version-4/edit-question/README.md
```diff
@@ -0,0 +1,91 @@
...
+- Users often needed to consider answer type before they were able to write useful hint text, but the current interface asks for hint text first  
+- They are also not shown how the answer type will display in the preview pane until they have saved a question. Some users re-wrote their hint text after saving and previewing  
+
+## What we changed and why
+
+- Removed the ‘short version’ input for a question. Users continued to query the need for it, and were unsure when or how they could use it. We decided that this is probably not needed as a basic feature, and if we return to consider it in the future it may be a more ‘advanced’ feature
+- Made the ‘hint text’ question and input always visible - no longer in a ‘details’ component - meaning it is more prominent
+- Added more context to the ‘save’ buttons making it clearer what action each will take
+- Made ‘save and add next question’ the main call to action when adding a new question to the end position in the form
+- Made ‘save and edit next question’ the main call to action when editing a question within the form journey, not last
```
C-Harry, line 57: Not totally clear about the 'not last' bit at the end of line 57: does this mean that when editing a question within the form journey and that question is NOT the last question, 'save and edit next question' becomes the main call to action? 

This suggests to me that when editing a question that IS the last question, something else will be the main call to action - 'Save and preview question' perhaps? But is that right?

christophercameron-ixd: It’s the latter - the bullet above is what the button (call to action) would be IF the current question the user is editing is the last one in their form.

C-Harry: So am I right that in this prototype version we now have the options of:

- 'Save and add next question' as the main CtA when a new question is added in the end position
- 'Save and edit next question' as the main CtA when editing a question that is NOT in the end position
- 'Save and preview' as the main CtA when editing a question that IS the last question? 

Whereas before we had 'Create new question' or 'Save question' as the main CtAs, respectively?

C-Harry (approved):
Have suggested a few minor amends to wording to help readability/comprehension

design/prototype-version-4/edit-question/README.md
christophercameron-ixd: So in the version (v4) the only green/main CtAs are either: 

- 'Save and add next question' as the main CtA when a new question is added in the end position
- 'Save and edit next question' as the main CtA when editing a question that is NOT in the end position

The ‘Save and preview question’ is always the same secondary (grey) button

C-Harry: Gottit :-)

design/prototype-version-4/README.md
```diff
@@ -225,27 +207,24 @@ Below is a secondary heading, “Declaration”, before the text “By submittin
...
+This round of testing saw a reduction in issues with creating and editing a question.  
+
+We continued to see problems when form creators were previewing their form in a new tab, with users being confused where they were and also changing ‘hats’ from form creator to form filler.  
 
-<!-- REMOVE ME!!! after write up -->
-<!-- https://drive.google.com/drive/folders/1VOMG9a0XyV371_1LxV9qUaddkyuY_xjE -->
+### User research documentation
 
-<br>
+[Usability testing: Basic form building write up (GitHub)](../../research/2022-06-28_Basic_Form_Building.md)
```
hannahkc, line 220: This link is a 404 - and I can't see what it should be linking to?

oliverquinlan: I can't see how to edit this from this view, but the link should be https://github.com/alphagov/forms/blob/main/research/2022-06-28_Basic_Form_Building.md

christophercameron-ixd: Ah yeah, it’ll go to where Oliver says, but it’ll only work when merged in because the filenames changed in Oliver’s recent rework. 
The link is relative to the current page/file location.

hannahkc:
ah ok cool!

hannahkc (approved):

PR #150, opened 2024-03-21
Update ADR019-use-amazon-ses.md

HollyCatling:
I've just updated to reflect the fact that we do now use Notify to send emails to members of the public.

# PR Checklist

---

# What

I've updated the ADR to reflect the fact that we now use Notify to send emails to memebers of the public.

# How to review

For example:

# Who can review


lfdebrux (approved):

james-sheppard-gov-uk:
It may be more appropriate to create an additional ADR specifically for sending confirmation emails. Let's discuss before we merge this.

chao-xian (changes requested):
I'll raise a new ADR

PR #153, opened 2024-03-28
ADR029 specify where we use Notify

chao-xian:
## Rendered view

https://github.com/alphagov/forms/blob/add-adr-29/ADR/ADR028-use-notify-for-users.md

## PR Checklist

---

# What

# How to review

For example:

# Who can review

### DIFF
--- a/ADR/ADR028-use-notify-for-users.md
+++ b/ADR/ADR028-use-notify-for-users.md
@@ -1,0 +1,30 @@
+# ADR029: Use GOV.UK Notify for emails sent to users for simple notifications
+
+Date: 2024-03-28
+
+## Status
+
+Accepted
+
+
+## Context
+
+This is an addendum to [ADR019](ADR019-user-amazon-ses.md) which set out the intention to use Amazon's Simple Email Service instead of GOV.UK Notify to send emails.
+
+GOV.UK Forms has continued to use GOV.UK Notify to send emails including:
+- submission confirmation emails to members of the public who have completed a form.
+- the submitted forms to the service processing them.
+- One Time Passcodes to verify the ownership of the submission email address when creating or editing a form.
+
+Amazon SES is currently used during the authentication process when logging into the form builder via Auth0, a third party authentication service, to send One Time Passcodes since Auth0 does not offer integration with the GOV.UK Notify API. Using Amazon SES also provides enforced TLS.
+
+GOV.UK Forms now needs to send notifications to users relating to account management (such as when they are upgraded from a trial to full user account).
+
+## Decision
+
+GOV.UK Forms will use GOV.UK Notify where the service is appropriate and does not require the additional features that Amazon SES offers.
+
+## Consequences
+
+- GOV.UK Forms does not need to invest in building additional monitoring, alert and administrative interfaces related to sending email which GOV.UK Notify already provide.
+- GOV.UK Forms will be dependent upon the support, reliability and feature roadmap of GOV.UK Notify, for example if and when they offer enforced TLS to the recipients inbox.


ADR/ADR028-use-notify-for-users.md
```diff
@@ -0,0 +1,20 @@
...
+Date: 2024-03-28
+
+## Status
+
+Accepted
+
+
+## Context
+
+Emails to users that confirm they have submitted their form, and as notifications for their user account management (such as when they are upgraded from a trial to full user account, need to be sent. We currently only use Amazon SES for submission emails where TLS requirements are higher. This is an addendum to [ADR019](ADR019-user-amazon-ses.md).
```
danworth, line 12: A closing bracket is missing. Also AFAIK we're currently only using AWS SES to send the Auth0 OTPs. [Submission emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_mailer.rb) and [submission confirmation emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_confirmation_mailer.rb) in runner and the [OTP to verify a change in submission email via admin](https://github.com/alphagov/forms-admin/blob/main/app/mailers/submission_email_mailer.rb) are sent via Notify aren't they?

chao-xian: Sorry I got confused and made this at the end of the week before Easter hol - will amend. Thanks!

chao-xian: odd because I tl;dr that Thursday afternoon 🫠

chao-xian: @danworth if it's ok, I've lifted your description

ADR/ADR028-use-notify-for-users.md
```diff
@@ -0,0 +1,20 @@
...
+Date: 2024-03-28
+
+## Status
+
+Accepted
+
+
+## Context
+
+[Submission emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_mailer.rb) and [submission confirmation emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_confirmation_mailer.rb) in runner and the [OTP to verify a change in submission email via admin](https://github.com/alphagov/forms-admin/blob/main/app/mailers/submission_email_mailer.rb) are sent via Notify. Notifications to users relating to account management (such as when they are upgraded from a trial to full user account), will be sent. We currently only use Amazon SES for Auth0 One Time Passcode emails where TLS requirements are higher. This is an addendum to [ADR019](ADR019-user-amazon-ses.md).
```
danworth, line 12: Nit picking but since these are externally viewable can we use `GOV.UK Notify` rather than just `Notify` (I know exactly what you mean). 

`, will be sent.` I presume this sentence should end with `using GOV.UK Notify`?

`emails where TLS requirements are higher` since Auth0 cannot integrate with Notify we never explored the option of using Notify for this and consequently whether not enforcing TLS was an issue, though its a benefit of using SES. It might be worth adding that using GOV.UK Notify was not an option for the Auth0 email flow.

chao-xian: NOT nitpicky! I'm not the best at ADR/RFC writing and really appreciate this. Have incorporated your suggestions.

ADR/ADR028-use-notify-for-users.md
```diff
@@ -0,0 +1,24 @@
...
+Date: 2024-03-28
+
+## Status
+
+Accepted
+
+
+## Context
+
+[Submission emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_mailer.rb) and [submission confirmation emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_confirmation_mailer.rb) in runner and the [OTP to verify a change in submission email via admin](https://github.com/alphagov/forms-admin/blob/main/app/mailers/submission_email_mailer.rb) are sent via GOV.UK Notify. Notifications to users relating to account management (such as when they are upgraded from a trial to full user account), will be sent using GOV.UK Notify.
```
danworth, line 12: I think we can remove the links to our source code since we're considering ADR immutable and those links could break with a bit of refactoring etc. Few minor suggestions, happy to discuss.

<blockquote>

This is an addendum to [ADR019](ADR019-user-amazon-ses.md) which set out the intention to use Amazon's Simple Email Service instead of GOV.UK Notify to send emails.

GOV.UK Forms has continued to use GOV.UK Notify to send emails including:
- submission confirmation emails to members of the public who have completed a form.
- the submitted forms to the service processing them. 
- One Time Passcodes to verify the ownership of the submission email address when creating or editing a form.

Amazon SES is currently used during the authentication process when logging into the form builder via Auth0, a third party authentication service, to send One Time Passcodes since Auth0 does not offer integration with the GOV.UK Notify API. Using Amazon SES also provides enforced TLS.

GOV.UK Forms now needs to send notifications to users relating to account management (such as when they are upgraded from a trial to full user account).
</blockquote>

ADR/ADR028-use-notify-for-users.md
```diff
@@ -0,0 +1,24 @@
...
+
+This is an addendum to [ADR019](ADR019-user-amazon-ses.md).
+
+## Decision
+
+We will use GOV.UK Notify where the service is appropriate and does not require the additional features that Amazon SES offers.
+
+## Consequences
+
+We keep using a free and simpler mechanism for sending emails to our users.
```
danworth, line 24: Just wondering in order to be balanced, whether its worth mentioning some of the following?
- GOV.UK Forms does not need to invest in building additional monitoring, alert and administrative interfaces related to sending email which GOV.UK Notify already provide.
- GOV.UK Forms will be dependent upon the support, maintenance and feature roadmap of GOV.UK Notify, for example if and when they offer enforced TLS to the recipients inbox.

ADR/ADR028-use-notify-for-users.md
```diff
@@ -0,0 +1,24 @@
...
+
+[Submission emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_mailer.rb) and [submission confirmation emails](https://github.com/alphagov/forms-runner/blob/main/app/mailers/form_submission_confirmation_mailer.rb) in runner and the [OTP to verify a change in submission email via admin](https://github.com/alphagov/forms-admin/blob/main/app/mailers/submission_email_mailer.rb) are sent via GOV.UK Notify. Notifications to users relating to account management (such as when they are upgraded from a trial to full user account), will be sent using GOV.UK Notify.
+
+We currently only use Amazon SES for Auth0 One Time Passcode emails, as we could not integrate GOV.UK Notify with Auth0 as they have a higher TLS requirement.
+
+This is an addendum to [ADR019](ADR019-user-amazon-ses.md).
+
+## Decision
+
+We will use GOV.UK Notify where the service is appropriate and does not require the additional features that Amazon SES offers.
```
danworth, line 20: I'm not sure if we have a style guide but I think using `GOV.UK Forms` instead of `we` makes sense?? I may very well have used `we` myself in the past...

DavidBiddle (approved):
This all makes sense to me, and it looks like you've incorporated all of Dan's feedback.

danworth (approved):
🏅

PR #164, opened 2024-05-17
Add ADR030: use of Lambda

AP-Hunt:
# PR Checklist

---

# What

We have begun to use AWS Lambda for running glue code. This ADR documents how we arrived at that decision.

# How to review

1. Copy review
2. Do we agree that this should be the path?

### DIFF
--- a/ADR/ADR030-use-aws-lambda-to-run-glue-code.md
+++ b/ADR/ADR030-use-aws-lambda-to-run-glue-code.md
@@ -1,0 +1,79 @@
+# ADR030: Use AWS Lambda to run glue code
+
+Date: 2024-05-17
+
+## Status
+
+> Accepted or Rejected
+>
+> Valid ADR statuses are: "Proposed", "Accepted", "Rejected", "Deprecated", "Superseded by ADRzzz", but since we use the GitHub Pull request mechanism to approve ADRs, "Proposed" isn't applicable.
+
+## Context
+
+There are, in this context, two types of code we want to consider
+
+1. Application code
+2. Glue code
+
+Application code is written by the developers on the team, and makes up the GOV.UK Forms service and all of
+its functionality. It is executed to handle HTTP requests received by a long-lived web server process, and it has access
+to stateful data storage like PostgreSQL and Redis.
+
+In contrast, glue code isn't part of any application, but is required to support the function of the service or its 
+infrastructure in some way. Examples of that could include code for integrating two independent systems, or code for 
+testing the state of something periodically.
+
+This ADR addresses where we run infrastructure code, and provides guidance on when not to follow the ADR.
+
+Our application code runs in containers on [AWS ECS on Fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Fargate.html).
+It is a managed service which is well suited to running long-running processes like web servers. It has a somewhat involved
+setup process and its use of containers incurs some maintenance and operational overhead, but that is a worthwhile
+trade-off for its functionality and stability. It is also able to run short-lived tasks.
+
+However, for infrastructure code, which is often characterized by being small, quick to run, and not requiring a connection to
+any stateful data stores, the overhead of the different pipelines and configurations required to run in ECS becomes
+disproportionate to the size of the code and the frequency with which we think about it. We can quickly end up with more
+code for running and configuring the thing than code in the thing we want to run.
+
+AWS Lambda is AWS' serverless functions-as-a-service service. It be given a source code bundle and it can execute the code within
+based on a small amount of configuration. Its primary target use case is in running code in response to events, without the
+need for a long running server. This matches up well with the types of use cases we have for glue code.
+
+## Decision: use AWS Lambda to run glue code
+
+We will use AWS Lambda for workloads that meet these criteria:
+* aren't application code
+* are glue code
+* are short-lived
+* are written in Ruby
+* are infrequently run (including on a schedule)
+* do not require access to persistent data storage
+* run asynchronously in response to some event
+* are not handling HTTP requests
+
+## When should I not use AWS Lambda?
+
+There are some glue code workloads that require access to persistent data stores, for example a daily job ensuring all 
+the users in our database are subscribed to our mailing list. Such a workload could be run in AWS Lambda, but we should 
+prefer running it in AWS ECS on Fargate. 
+
+We should prefer it because we already have secure connections to the data stores configured, and the code can be implemented
+as a Rake task to benefit from the available application logic. The workload itself can use the same container image as
+the application, which reduces the maintenance overhead.
+
+We should strictly avoid using AWS Lambda for any workload that handles HTTP requests. The additional complexity and costs
+(financial, organisational, and human) in our opinion massively outweigh any positives AWS Lambda has to offer.
+
+## Consequences
+
+The negative consequence of choosing AWS Lambda for certain types of workloads is that we have a second compute service in use 
+within our infrastructure, which comes with the maintenance overhead associated with adopting any new service. 
+
+Having a second compute service in use forever raises the question of where any given piece of code should run, however
+we hope to mitigate that problem with the guidance given in this ADR.
+
+We hope to mitigate the maintenance overheads of using AWS Lambda, such as updating language runtimes and dependency management, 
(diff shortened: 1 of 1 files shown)


sarahseewhy (approved):
Accepted

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+> Valid ADR statuses are: "Proposed", "Accepted", "Rejected", "Deprecated", "Superseded by ADRzzz", but since we use the GitHub Pull request mechanism to approve ADRs, "Proposed" isn't applicable.
+
+## Context
+
+There are, in this context, two types of code we want to consider
+
+1. Application code
+2. Glue code
+
+Application code is that which is written by the developers on the team, and makes up the GOV.UK Forms service and all of
```
sarahseewhy, line 18: Try, _Application code is written by developers_. It's a tad more active voice.

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+its functionality. It is executed to handle HTTP requests received by a long-lived web server process, and it has access
+to stateful data storage like PostgreSQL and Redis.
+
+Glue code is the code isn't part of any application, but is required to support  the function of the service or its
```
sarahseewhy, line 22: Similarly try _In contrast, glue code isn't part of any application_. Leaning into the active voice.

AP-Hunt: Done :+1:

ADR/ADR030-use-of-aws-lambda.md
AP-Hunt: Done :+1:

chao-xian:
My 2 pennies worth that isn't a stopper: I am generally not convinced by lambdas/serverless as I've only heard how the hassle of running/maintaining outweighs their usefulness (cannot remember details). As long as it doesn't impact devs (from a really selfish pov) on the team then I'm not too against it. However @danworth seemed to indicate a stronger view.

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+trade-off for its functionality and stability. It is also able to run short-lived tasks.
+
+However, for infrastructure code, which is often characterized by being small, quick to run, and not requiring a connection to
+any stateful data stores, the overhead of the different pipelines and configurations required to run in ECS becomes 
+disproportionate to the size of the code and the frequency with which we think about it. We can quickly end up with more
+code for running and configuring the thing than code in the thing we want to run. 
+
+AWS Lambda is AWS' serverless functions-as-a-service service. It be given a source code bundle and it can execute the code within 
+based on a small amount of configuration. Its primary target use case is in running code in response to events, without the
+need for a long running server. This matches up well with the types of use cases we have for glue code.
```
DavidBiddle, line 42: I think a lot of this would make more sense in the Context section, so that the Decision section is strictly limited to the specifics of the decision rather than the reasoning for it

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+
+We should use AWS Lambda for workloads that meet these criteria:
```
DavidBiddle, line 44: Suggested language tweak to make it sound more decision-y:
```suggestion
We will use AWS Lambda for workloads that meet these criteria:
```

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+* aren't application code
+* are glue code
+* are short-lived
+* are infrequently run (including on a schedule)
```
DavidBiddle, line 48: We could cover the data storage point from below here briefly so that it's obvious at a glance:

```suggestion
* are infrequently run (including on a schedule)
* do not require access to persistent data storage
```

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,73 @@
...
+
```
danworth, line 49: As per David's point you cover the following in the introduction and in the "glue code" vs "application code" but to make it clear in this list I would suggest adding something about the following, though you may choose to word it better and if its not clear what I'm getting at please just let me know.
- Process asynchronously to the triggering event 
- Are not triggered by an HTTP request

With the above I'm trying to make it clear that this ADR does not include using lambdas for browser and web based requests, but simple asynchronous event based activities.

ADR/ADR030-use-of-aws-lambda.md
danworth, line 44: Can we add 'written in Ruby' to this please to avoid further divergence of languages within the team and the temptation to use other languages supported by lambda.

danworth (commented):
This is great to document the use cases for Lambda, and of course it can change over time but being clear now is very useful 🏅 

I've one or two comments for your review mostly born out of an abundance of caution. In addition to them, I'm wondering if this ADR is the right place to state that all Lambda logs should go to Splunk, the concern being that if they don't then these small glue code lambdas become less visible and harder for the uninitiated to work with. We could of course have a separate ADR stating what should and shouldn't be in Splunk.

AP-Hunt:
> My 2 pennies worth that isn't a stopper: I am generally not convinced by lambdas/serverless as I've only heard how the hassle of running/maintaining outweighs their usefulness (cannot remember details). As long as it doesn't impact devs (from a really selfish pov) on the team then I'm not too against it.

I'm 100% with you on this. I don't want our Lambda usage to bleed into any part of the service that faces a user. In an ideal world, AWS Lambda would only ever be thought about by SREs. Code written by devs should continue to run in containers in ECS.

ADR/ADR030-use-of-aws-lambda.md
AP-Hunt: Good idea. I've moved it.

ADR/ADR030-use-of-aws-lambda.md
AP-Hunt: Added

ADR/ADR030-use-of-aws-lambda.md
AP-Hunt: I've added the bullets you suggested. I've also added this to the "When should I not use AWS Lambda" section

> We should strictly avoid using AWS Lambda for any workload that handles HTTP requests. The additional complexity and costs
(financial, organisational, and human) in our opinion massively outweigh any positives AWS Lambda has to offer.

DavidBiddle (approved):
Thanks for making those changes - I think this looks good stylistically, and the rules you've set down seem sensible and pragmatic to me.

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,79 @@
+# ADR030: Use AWS Lambda small tasks run infrequently
```
james-sheppard-gov-uk, line 1: Perhaps call this ADR "Use AWS Lambda to run glue code"

ADR/ADR030-use-of-aws-lambda.md
```diff
@@ -0,0 +1,79 @@
...
+However, for infrastructure code, which is often characterized by being small, quick to run, and not requiring a connection to
+any stateful data stores, the overhead of the different pipelines and configurations required to run in ECS becomes
+disproportionate to the size of the code and the frequency with which we think about it. We can quickly end up with more
+code for running and configuring the thing than code in the thing we want to run.
+
+AWS Lambda is AWS' serverless functions-as-a-service service. It be given a source code bundle and it can execute the code within
+based on a small amount of configuration. Its primary target use case is in running code in response to events, without the
+need for a long running server. This matches up well with the types of use cases we have for glue code.
+
+## Decision: use AWS Lambda
```
james-sheppard-gov-uk, line 42: Use AWS Lambda for glue code

james-sheppard-gov-uk:
Some very minor comments from me about being clear it's for "glue code" as part of the title and decision. I suggest the filename is also changed to `ADR030-use-aws-lambda-for-glue-code.md`.

I support the use of AWS Lambda for the right use cases!

AP-Hunt:
Addressed @james-sheppard-gov-uk's comments

chao-xian (approved):

Issue #182, opened 2024-11-08
License?


bzg:
Thanks for sharing this code!

I see several repositories from alphagov are published under a Free Software license.

Will it be the case for this repository as well?

chao-xian:
Thanks @bzg - we've added a licence.

PR #183, opened 2024-11-11
Add licenses

chao-xian:
# What

Adding a missing LICENSE file. Also add link to the Open Government License.

Fixes https://github.com/alphagov/forms/issues/182

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2022 Crown Copyright (Government Digital Service)
+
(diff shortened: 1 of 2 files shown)


LICENSE
lfdebrux, line 1: GDS Way recommends we name the file `LICENCE` as that's British English: https://gds-way.digital.cabinet-office.gov.uk/manuals/licensing.html#specifying-the-licence

chao-xian: d'oh - I always get those mixed up (and need to correct the files in the other repos in that case!)

PR #186, opened 2024-11-25
Add ADR 038 - Store submission data in PostgreSQL and use Solid Queue to send emails asynchronously

chao-xian:
[Rendered view of the ADR](https://github.com/alphagov/forms/blob/add-adr-36/ADR/ADR036-add-solidqueue-to-forms-runner.md)

# PR Checklist

---

# What

ADR for adding SolidQueue

# How to review

For example:

# Who can review

@forms-devs @james-sheppard-gov-uk  @AP-Hunt @cadmiumcat @alice-carr


ADR/ADR036-add-solidqueue-to-forms-runner.md
```diff
@@ -0,0 +1,37 @@
...
+
+Date: 2024-11-25
+
+## Status
+
+Approved
+
+## Context
+
+Forms will be using AWS SES to deliver emails to form processors with files attached to the emails. For processing of these emails, we looked at Sidekiq and Redis for managing a queue and to give us the ability to retry failed submissions. However with Redis, the details of a submission are ephemeral once a job is taken to be processed. Due to the way Sidekiq uses Redis in the free version is it possible for it to lose jobs if the process crashes. It is very important that we do not lose submissions, so this risk isn't acceptable to us.
```
cadmiumcat, line 11: To understand the context better it would be helpful to have a sentence or list that explains what our requirements are. Then we can explaining the options we considered and why they do/don't work. For example:

> For our team to successfully process submission emails we must be able to:
> - retry sending the email on failed submissions
> - a queuing mechanism to send emails asynchronously  
> - no limit to message size in the queuing system
(I don't know if that's what we need... but I think that's why I'd find it clearer)

ADR/ADR036-add-solidqueue-to-forms-runner.md
```diff
@@ -0,0 +1,37 @@
...
+
+We considered just using SQS, but there is a message size limit of 256kb. We cannot be sure that there won’t be form submissions that exceed this size.
```
cadmiumcat, line 13: Will we ever have a maximum size, or do we allow any size?

cadmiumcat: There seem to be [workarounds that involve putting the message payload in s3](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-s3-messages.html), rather than the queue itself. The queue contains a reference to the object in s3. It seems to suggest this is up to a limit of 2GB

AP-Hunt:
I have some concerns about using a database as a queue. They fall in three buckets: performance,  restoration, and backup retention.

## Concerns

### Performance

Using the same database instance for `forms-admin`, `forms-api`, AND `forms-runner` will naturally place additional load on that database. This is more than likely OK during low and average traffic times, but I’m worried about the impact during high load times.

The three apps using the same database instance means their performance becomes interconnected. One app seeing increased request volumes places increasd load on the database, which can degrade the performance of the other apps as the database slows. 

I can immediately think of a scenario where a form is receiving a huge volume of submissions, resulting in huge volumes of jobs on the queue and increased database activity to process them all. This could slow down the responses from `forms-api`, on which `forms-runner` depends to render a form to begin with. 

To my knowledge, the web server we use, Puma, is multi-threaded but not asynchronous. This means that a single thread can only handle a single request at a time, and will be blocked on IO operations such as waiting for a response to its own HTTP request. The increased response time from `forms-api` locks up Puma’s threads for longer and longer, which slows down overall response time, and increases the size of whatever request queue Puma may have. This in turn would result in (severely) degraded performance of the web side of `forms-runner`  (assuming job processing will be a different set of containers).

We think we saw _some_ of this kind of behaviour when we were performing load tests in late 2023, specifically where `forms-api` response times slowed and had a knock on effect on `forms-runner`.

Some of these potential performance pitfalls might be mitigated somewhat by our use of AWS Aurora, which can auto-scale. However, right now, today, we have our minimum and maximum scaling parameters set to the same thing. I don’t think we’ve ever tested the speed at which Aurora can scale, or what our limits should be.

### Restoration

We perform database backup and restoration operations at the database instance level. This means that we’re operating on every database within the instance all at once ; when we want to backup `forms-admin` database we also backup `forms-api` database, and when we want to restore/roll back `forms-admin` database we also restore/roll back `forms-api` database.

We already acknowledge that this is a problem for us, because in a rollback situation we will lose any data between now and the time we’re restoring to in _both_ databases, and this is typically undesirable. If we did that, we’d have to find some way of also restoring the lost data.

If we add a database-as-a-queue to that same database instance we make this problem potentially much worse because we’ll risk losing user submissions too.

### Backup retention

We [retain our database backups for 30 days](https://github.com/alphagov/forms-deploy/blob/main/infra/deployments/forms/tfvars/production.tfvars#L16). This is good and proper and meets our needs.

In our current setup, user submissions are only held in Redis for a maximum of 24 hours, and [we do not take any backups](https://github.com/alphagov/forms-deploy/blob/main/infra/modules/redis/elasticache.tf#L26) of it. That means we never store user submissions for longer than 24 hours.

If we begin using the same database instance for the job queue containing user submissions, a subset of those user submissions will be captured in our nightly backups and stored for 30 days. I’ve read our privacy policy and it sounds like that might be ok, but probably not desirable. I don’t know if our MOU is any more stringent.

## Suggestions

I’d like to suggest we reconsider the use of AWS SQS.  AWS SQS does have an upper limit on message size of 256Kb, but that is a LOT of JSON. For example, here is [a 256Kb blob of minified JSON](https://microsoftedge.github.io/Demos/json-dummy-data/256KB-min.json). It contains 231,797 characters.  

To put it in more perspective, l’ll try to show how many words that is.

Assume we’re using UTF-8 encoding. UTF-8 uses between 1 and 4 bytes for each character, so we’ll assume the worst case and say a UTF-8 character is 4 bytes.

Let’s also estimate that the average English word is 5 letters, 6 with a space after it.

With 256,000 bytes to use, that’s `256000/(6*4) = 256000/24 = 10,666` words. For comparison, a Bachelors degree dissertation is somewhere in the region of 10k words.

As a very rough validation, I asked [Lipsum.com](https://www.lipsum.com/) for 256kb of filler text. It gave me 100Kb (🤷) consisting of 14,791 words 

image

Of course those numbers are all for English words, and not JSON payloads. However, hopefully you get the picture: there’s loads of room.

### Extended SQS message pattern

If message sizes above 256kb are still a concern for us, there are workarounds.

The upper limit on SQS message size is a well known problem with [a well known solution](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-s3-messages.html). In short, write the full payload to S3 and write metadata to the queue. Unfortunately there’s no AWS-supported extended SQS client for Ruby, but the thing it does shouldn’t be too hard to re-implement for ourselves. 

We can set up AWS S3 object lifecycle policies to ensure messages are deleted from S3 after N days if they fail to be deleted by the job processor.

ADR/ADR036-add-solidqueue-to-forms-runner.md
```diff
@@ -0,0 +1,38 @@
...
+Date: 2024-11-25
+
+## Status
+
+Approved
+
+## Context
+
+Forms will be using AWS SES to deliver emails to form processors with files attached to the emails. For processing of these emails, we looked at Sidekiq and Redis for managing a queue and to give us the ability to retry failed submissions - such as bounced emails. However with Redis, the details of a submission are ephemeral once a job is taken to be processed.
+Due to the way Sidekiq uses Redis in the free version is it possible for it to lose jobs if the process crashes. It is very important that we do not lose submissions, so this risk isn't acceptable to us.
```
DavidBiddle, line 12: It might make sense to link out to [Sidekiq's reliability page](https://github.com/sidekiq/sidekiq/wiki/Reliability) here to make it easier for people to check whether it's still true in the future

ADR/ADR036-add-solidqueue-to-forms-runner.md
stephencdaly: Yeh I think this would be helpful. In particular we need both of the things you mentioned:

- A queueing mechanism to send emails asynchronously
- A way to retry the email on failed submissions, for example if an email bounces. 

In particular with bounced emails, we only hear about this some time after the email attempt. A notification is published to an SNS topic that we will subscribe an SQS queue to. Ideally we would automatically retry a number of times, but we will need to keep the data until we have successfully managed to send it on (within whatever limits our policies allow).

The message size limit requirement is slightly uncertain. If we think, for example, the SQS limit is fine for now - we'd like to know if it ever becomes not fine, and ideally not lose any user data in this case.

stephencdaly:
@AP-Hunt On the performance front, whilst I think we should be careful about this, we should have ways to manage this. We could limit the number of worker processes reading from the queue so that even if we have a high volume of submissions we wouldn't be processing more than we can handle - of course we'd have to make sure we have enough worker processes so as not to build up a too big backlog. We should probably run some performance tests to tune this and have confidence.

The thing that's more of a concern I think, as you say, is the database backups. We should probably explore the S3 option you suggested - this should work with giving us the ability to retry emails if they bounce too. If we still think the database is the best solution we'd have to work with IA people to see if this okay, but I think we would prefer to store submission data for as little time as possible.

ADR/ADR036-add-solidqueue-to-forms-runner.md
stephencdaly: Nice, thanks for this suggestion! I think enforcing a size limit would be difficult - I guess we could add a maximum to free text fields. This probably would have been 8kb per question due to the WAF rules, but now we're letting through larger requests. So theoretically forms could ask for essay-length submissions although I imagine this is very unlikely. So it probably just depends on whether we're happy with that small risk of losing user submissions if they're too large.

I think what is more important, and a reason we probably can't put the submission data just in the SQS message is that the email might bounce - which we hear about later. In this case we need to make sure we're able to retry sending the submission email, so we'd want the data to be maintained until we know it's been successfully delivered.

With the S3 solution you suggested, I think this should be possible. We'd just need a way of linking an email send attempt back to the data in S3 so we can delete the file when it's successful, or retry if it's unsuccessful. Possibly we can use the message ID for this and have it the same as the s3 key - but we'd need to check how this would work.

cadmiumcat: Yeah, and I think Andy's calculations show that we'd still be quite unlikely to hit the SQS limit anyway. Unless there are weird use cases we haven't considered

DavidBiddle: I remember the engagement team mentioning an incredibly long form (100+ page PDF) for setting up a school and I had that in mind as a possible problem form. But that's a very unusual case and I'm not even sure that would hit the limit

cadmiumcat:
@chao-xian there's already an ADR036. And an open ADR037

chao-xian:
From discussion with @james-sheppard-gov-uk, @AP-Hunt noted that we will need a database for Runner in the future for storing other kinds of metadata for forms and submissions. To overcome the issues with performance, we've agreed that giving Runner it's own database instance gets around most of the concerns.

ADR/ADR036-add-solidqueue-to-forms-runner.md
chao-xian: Have incorporated some of this! Thank you

ADR/ADR036-add-solidqueue-to-forms-runner.md
chao-xian: I'm not sure it's worthwhile doing this here...

ADR/ADR038-add-solidqueue-to-forms-runner.md
```diff
@@ -0,0 +1,46 @@
...
+Approved
+
+## Context
+
+Forms will be using AWS SES to deliver emails to form processors with files attached to the emails.
+
+For our team to successfully process submission emails we must have:
+* a way to retry sending the email on failed submissions
+* a queuing mechanism to send emails asynchronously
+* a way to retry the email on failed submissions, for example if an email bounces.
```
james-sheppard-gov-uk, line 16: How is this different to "a way to retry sending the email on failed submissions" on line 14?

ADR/ADR038-add-solidqueue-to-forms-runner.md
```diff
@@ -0,0 +1,46 @@
...
+
+We will use SolidQueue and ActiveJob to manage the background processing of emails. We will use PostgresSQL/RDS for the database on the Forms Runner.
+
+## Consequences
+
+We will have a reliable way to deliver emails to form processors with files attached to the emails.
+
+We must encrypt the data in the database, as this will be user-submitted data.
+
+We have an additional infrastructure dependency on PostgresSQL on Forms Runner.
```
james-sheppard-gov-uk, line 46: From my understanding this requires an additional Aurora cluster. The additional cluster might be worth capturing as a new ADR, superseding https://github.com/alphagov/forms/blob/main/ADR/ADR010-rds-configuration.md

chao-xian: That's a good shout I'll do that

ADR/ADR038-add-solidqueue-to-forms-runner.md
chao-xian: Ah good spot - this was a copy paste combination of things two of us wrote

chao-xian:
Regarding https://github.com/alphagov/forms/pull/186#discussion_r1888430549 I'll start that immediately after this one is accepted/merged

ADR/ADR038-add-solidqueue-to-forms-runner.md
chao-xian: @AP-Hunt and @james-sheppard-gov-uk I've added:
> In terms of performance and additional infrastructure, we have plans for the future to store form filling data in something more persistent than Redis as part of things like Save and Return. Given that we will need that resource anyway, setting up an additional database for Forms Runner isn't a problem. Another ADR will be forthcoming separately to record that we will need an additional Aurora cluster, but worth mentioning here as it's relevant.

jamie-o-wilkinson:
It seems to me like there are two separate main parts to this:
-  we need to store submission data somewhere, to allow retries if SES email delivery fails/bounces (and also for future features like save and return)
- we need a way of queuing submissions

I thought this was just about the queuing, so was confused why weren't just thinking of using SQS, as it would give us things like a dead letter queue, retries, and manual redrives out of the box without necessitating a new database. I understand dealing with email bounces would make this tricky as as far as the SQS queue is concerned, a submission with a bounced email was dealt with "fine" and the submission data would be lost. We'd need to subscribe to the success SNS topic and only clear messages then. 

I'm not yet certain that SQS couldn't do what we need though...

chao-xian:
@jamie-o-wilkinson thanks Jamie that's a good point. The problem we were focusing on is:

> Due to the way Sidekiq uses Redis in the free version is it possible for it to lose jobs if the process crashes. It is very important that we do not lose submissions, so this risk isn't acceptable to us.

I'm not sure SQS doesn't have this issue, but can't tell from reading through the docs atm

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
```diff
@@ -0,0 +1,90 @@
...
+
+As we need to store data in a database to [handle bounced emails](#storing-submission-data-to-handle-bounced-emails), the message size limit of 256 KB will not be an issue for us. We can just send an ID referencing the submission in the SQS message. The cost associated with sending the number of SQS messages that we anticipate is negligible.
+
+There is an AWS gem, `aws-activejob-sqs-ruby` to integrate SQS with ActiveJob in Rails. However, this is not widely used.
+
+#### Chosen technology
+
+We have decided to use ActiveJob backed by Solid Queue to send emails asynchronously. This is because this is the default way of processing background jobs in Rails and we cannot see significant advantages to using SQS. As we require a SQL database for storing submission data, needing a database is not a blocker to using Solid Queue.
```
cadmiumcat, line 62: I don't particularly have a problem with ActiveJob + Solid Queue, but given that we've said "lots of people on the team are familiar with" SQS, it seems a bit odd that we wouldn't prefer that.

Are there any other reason we would rather go with this than AWS SQS?

chao-xian: One particular reason is that it's standard in Rails now, and Rails itself manages some moving parts for us such as inherently knowing which jobs are related to which submissions in our system.

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
```diff
@@ -0,0 +1,90 @@
...
+
+We will use PostgreSQL/RDS for the database on the Forms Runner, with a separate Aurora instance.
+
+## Consequences
+
+We will have a reliable way to deliver emails to form processors with files attached to the emails.
+
+Forms Runner will be responsible for running background jobs to send submission emails and consume SQS messages that inform us about the send result. We will have to manage any performance impacts of this and configure additional monitoring.
+
+We have an additional infrastructure dependency on PostgreSQL for Forms Runner. There will be additional costs to configuring another Amazon Aurora instance.
```
cadmiumcat, line 90: Something in the back of our (infra) minds is RDS blue/green deployments. We want to avoid downtime with our current database, but we haven't had the capacity to work on this.
Having a database that submissions rely on probably increases the need for blue/green deployments

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
cadmiumcat: 👍 I think it's worth adding it here then as a reason to prefer it

stephencdaly: Yeh I did struggle a bit with pinning down exactly why we want to use Solid Queue over SQS - I think it's close. I think the preference for Solid Queue comes from the fact it is the Rails default, and is well supported for use with ActiveJob which makes it very easy to schedule tasks in the code.
SQS does work with ActiveJob but the gem for it only has [~25,000 total downloads](https://rubygems.org/gems/aws-activejob-sqs/), which is tiny which would make me hesitant about using it.

I tried to look for comparisons of using SQS vs ActiveJob but I can't really find people talking about that comparison.

Personally, I'm still open to us using SQS if it gives us things that Solid Queue doesn't but I'd also be quite keen on using the Rails default and keeping everything in the database unless we see glaring issues with it. As long as we can get the monitoring for it in place, using the thing that "just works" seems like a good idea.

I did just think of another advantage of Solid Queue which should probably be added in, which is that there probably isn't much setup to get it working locally for testing.

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
stephencdaly: Thanks, I can add that to the consequences

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
cadmiumcat: Thanks, I thought I was missing something obvious, but that fits in with my understanding. 
My gut feeling says "why would you introduce another thing when we're already using SQS?". But I also see the advantage of using a thing that just works, particularly if it makes it easier for devs to do testing.

My comments are definitely not blocking

chao-xian: Yes there's certainly nothing majorly restricting us from switching to SQS if we needed to in the future.

jamie-o-wilkinson: I feel similarly conflicted over Solid Queue vs SQS. My instinct to follow convention is hampered as Rails' convention to use Solid Queue conflicts with the convention to just use SQS for queueing regardless of framework, because it is a de facto standard (providing you use AWS).

I suppose from the code's point of view they should be indistinguishable, as they just use ActiveJob. So perhaps it comes down to what we think is going to be less overall work for the team?:
1. Infrastructure for and configuration of PostgreSQL, Solid Queue, and Active Job
2. Infrastructure for and configuration of SQS and Active Job

stephencdaly: I've tried to list out the pros and cons I can think of for Solid Queue and SQS to add in, can anyone think of anything else for either?

For Solid queue:

##### Pros

* Solid Queue is the Rails default, and is well supported
* It's easy to run Solid Queue on a developer machine for local testing
* It's easy to configure recurring tasks. We think we'll need to run a recurring task to delete submission data when our retention period has elapsed.

##### Cons

* We need to put more thought into how we add useful monitoring and alerting
* Extra configuration is required to set up a job management dashboard for observability

For SQS:

##### Pros

* We already use SQS for bounce notifications for SES emails
* Lots of the team are familiar with SQS
* We know how to set up alerts for messages going into the dead letter queue, and if the queue gets too big.
* Manual recovery from small numbers of failed submissions should be fairly straightforward - we can add messages to the queue from the AWS dashboard.

##### Cons

* The Ruby gem to use ActiveJob to queue tasks using SQS isn't widely used
* To test running background jobs on a developer machine we'll need additional setup to connect to AWS or run an SQS mock locally

chao-xian: Something about fewer moving parts to configure and manage with SolidQueue? Also easier to switch from SolidQueue -> SQS in future than the other way round?

stephencdaly: Have added the pros and cons and a caveat to the decision that we are open to switching to SQS

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
stephencdaly: Have added something about this.

ADR/ADR038-add-psql-database-and-solid-queue-to-forms-runner.md
cadmiumcat: Thanks so much for breaking it down

cadmiumcat (approved):

chao-xian:
TYSM @stephencdaly for getting this over the line for me 👏

PR #198, opened 2025-02-12
ADR040: Make container image tags mutable

AP-Hunt:
# PR Checklist

---

# What

We need to enable tag mutability in ECR for review apps to work.

# How to review

Copy review. Do you agree? Not agree? Do I need to add anything else?

### DIFF
--- a/ADR/ADR040-make-container-image-tags-mutable.md
+++ b/ADR/ADR040-make-container-image-tags-mutable.md
@@ -1,0 +1,45 @@
+# ADR040: Make container image tags mutable
+
+Date: 2025-02-12
+
+## Status
+
+> Accepted or Rejected
+>
+> Valid ADR statuses are: "Proposed", "Accepted", "Rejected", "Deprecated", "Superseded by ADRzzz", but since we use the GitHub Pull request mechanism to approve ADRs, "Proposed" isn't applicable.
+
+## Context
+
+At the time of writing, we use [Amazon ECR](https://aws.amazon.com/ecr/) to host our container image repositories. 
+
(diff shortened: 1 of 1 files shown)


chao-xian (approved):

ADR/ADR040-make-container-image-tags-mutable.md
```diff
@@ -0,0 +1,45 @@
...
+
+We are implementing review applications for `forms-admin` (see [ADR037](./ADR037-implement-review-apps.md)). In our implementation we want to deploy the version of `forms-admin` under review alongside the latest production version of `forms-api`. Deploying the latest production version means two things:
+
+1. we don't need to constantly update the code to change the version being deployed
+2. we need a mechanism for consistently identifying the latest production version
+
+The `latest` tag is conventionally the tag used to denote the most recent version of a container image, or otherwise the latest production-ready image. We see no reason not to follow convention, and have begun [tagging the newest production-ready images](https://github.com/alphagov/forms-deploy/pull/1346).
```
theseanything, line 24: Nit: How are we adding the "latest" tag? In a lot of cases "latest" means most recently pushed image to the registry, rather than most "recent version".

AP-Hunt: The PR linked there (`forms-deploy` #1346) has the implementation. But in short, we have a flag set in the production container deploying pipelines which add the latest tag at the same time they add the tag which means "this ran fine in production". None of those pipelines can have two run happens simultaneously; that is to say two deployments of `forms-admin` can't be happening concurrently, but `forms-admin` and `forms-api` can be deployed at the same time.

theseanything (approved):

PR #199, opened 2025-02-12
Add ADR for ActiveRecord Encryption

chao-xian:
# PR Checklist

---

# What

# How to review

For example:

# Who can review

### DIFF
--- a/ADR/ADR041-use-activerecord-encryption.md
+++ b/ADR/ADR041-use-activerecord-encryption.md
@@ -1,0 +1,37 @@
+# ADR041: Use ActiveRecord encryption in Runner
+
+Date: 2025-02-12
+
+## Status
+
+Accepted
+
+## Context
+
+Runner will be storing user submitted data which is likely to include sensitive information, such as PII including document IDs like passport numbers.
+
+Aurora RDS databases will already encrypt when data is saved to disk with AWS KMS, but while data is held in memory it is not. To mitigate against this we can additionally encrypt at the application level, so data is not stored in the database in plain text. Though a low risk, accidental leakage is doubly assured against having that sensitive user information exposed.
+
+Rails makes this quite easy to do with [ActiveRecord Encryption](https://guides.rubyonrails.org/active_record_encryption.html).
+
+### Pros
+
+* If there is more access to credentials to query data in the database than there is to your encryption keys. For example, if we allowed devs with the support role in AWS to query the database, we might want to make sure they don’t have access to user submission data. However, if we decide to do this in the future we can start encrypting the data later, as Active Record encryptions supports reading unencrypted data when encryption has been enabled.
+* If sensitive data is leaked, for example in the SQL logs, it's encrypted and unreadable.
+* An additional layer of defence from an accidental information disclosure by our engineers who have access to the database. Instances such as an engineer being able to read the plaintext entries while querying the database, or that information being accidentally saved to logs or out to monitoring systems.
+* Mitigates against leaks or modification if restrictions to our database backups is compromised, say from leaked credentials. Though it is very low likelihood that the access encryption key that has access to both the backups and the database encryption would also compromised.
+* Mitigates against internal malicious actors being able to read or modify data. However we already require security clearance, so impact is minimal unless we restrict access to the keys and the app itself.
+* This additional layer is something we can assure departments with so they have more confidence in using Forms.
+
+### Cons
+
+* Extra keys to manage, extra configuration in Rails.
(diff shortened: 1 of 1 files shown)


ADR/ADR041-use-activerecord-encryption.md
```diff
@@ -0,0 +1,32 @@
...
+### Pros
+
+* If there is more access to credentials to query data in the database than there is to your encryption keys. For example, if we allowed devs with the support role in AWS to query the database, we might want to make sure they don’t have access to user submission data. However, if we decide to do this in the future we can start encrypting the data later, as Active Record encryptions supports reading unencrypted data when encryption has been enabled.
+* If sensitive data is leaked, for example in the SQL logs, it's encrypted and unreadable.
+* An additional layer of defence.
+
+### Cons
+
+* Extra keys to manage, extra configuration in Rails.
+* Difficulty accessing the data if we ever need to inspect it. The main reason I can see for doing this is if for some reason we fail to send a submission due to a data error and need to inspect the data itself to diagnose and fix any issues.
```
AP-Hunt, line 24: This could be mitigated by having a rake task, an audit log, and an alert.

stephencdaly: If we have a rake task to inspect the data, is there any point in encrypting it? Only devs with admin access can currently even query the database. By adding encryption at the application level, are we doing anything more than preventing people with AWS admin access from reading the data? These people would still be able to obtain the key we use for encryption, so are we doing anything more than slowing them down slightly and making things more difficult for us?

ADR/ADR041-use-activerecord-encryption.md
```diff
@@ -0,0 +1,32 @@
+# ADR041: Use ActiveRecord encryption in Runner
```
theseanything, line 1: Is this ADR about whether we encrypt data at the application? or have we've already agreed that, and this is this specifically about implementation details and whether we use ActiveRecord encryption?

theseanything (approved):

chao-xian:
Good to get people's perspectives on this and it's definitely leading my mind more towards "let's not bother". I think I came from a position of being cautious by default but as I can't define an additional benefit of it, and others have said there doesn't seem to be clear risks that aren't already mitigated for, I'm happy to closed this PR or mark the ADR as rejected so we have a clear record for that decision. What are your thoughts on the latter @theseanything ?

theseanything:
So my thinking is, this would help us mitigate against:
- Accidental information disclosure from our engineers who have access to the database. For example, being able to read to the plaintext entries when querying the database. Or being saved in logs or monitoring systems.
- Information disclosure or modification if access to backups or the database is compromised (e.g. leaked credentials.). This depends on the likelihood that access encryption key also being compromised is low.  (I.e. credentials with permissions to access both).
- Internal malicious actors reading, leaking or modifying data. This is already mitigated somewhat by our security clearance requirements. The impact of this is minimal unless we restrict access to the keys and the application itself.

The reducing accidental disclosure or risks during compromised access are big enough benefits to do it, compared to the drawbacks described as:
- key management can be somewhat simplified if we use services like KMS 
- we can inspect data via the application (though we should have a way to audit access to data).

Are there any other drawbacks?  Performance overheads? Restriction in functionality i.e. [not being able to query the database](https://guides.rubyonrails.org/active_record_encryption.html#deterministic-and-non-deterministic-encryption), is deterministic encryption suitable?

chao-xian:
Thanks @theseanything that's super useful and I'll add it to the doc

chao-xian:
@theseanything have updated with your main points

ADR/ADR041-use-activerecord-encryption.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Runner will be storing user submitted data which is likely to include sensitive information, such as PII including document IDs like passport numbers.
+
+Aurora RDS databases will already encrypt when data is saved to disk with AWS KMS, but while data is held in memory it is not. To mitigate against this we can make use [ActiveRecord Encryption](https://guides.rubyonrails.org/active_record_encryption.html).
```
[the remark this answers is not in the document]
theseanything: Kelv has updated the ADR with them.

ADR/ADR041-use-activerecord-encryption.md
chao-xian: Have reframed to be about encrypting data at application level, and that the way to do it would be via ActiveRecord Encryption.

PR #222, opened 2025-08-01
Create ADR for proposed Welsh forms API implementation

DavidBiddle:
# PR Checklist

---

# What

# How to review

For example:

# Who can review


DavidBiddle:
Spike branches for option 1:
-  forms-api: https://github.com/alphagov/forms-api/tree/spike-add-welsh-columns-to-api-snapshot
- forms-runner: https://github.com/alphagov/forms-runner/tree/read-translated-fields-from-json

Spike branches for option 2:

- forms-api: https://github.com/alphagov/forms-api/tree/spike-translate-snapshot-at-request-time
- forms-runner: https://github.com/alphagov/forms-runner/tree/request-language-from-api

Both spikes use a rebased version of [the original Mobility spike branch](https://github.com/alphagov/forms-api/pull/772) as a base.

There's no spike branch for option 3 - I thought of this after I'd investigated and started writing up the other two, and it's a lot more complex to look into. If people think it's a better approach than the one proposed in this ADR, we can make a spike branch to investigate it further next week. The runner implementation would be the same aas option 2, but I think we'd have to make quite a few changes in admin and API.

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,336 @@
...
+  ]
+}
+```
+
+Advantages:
+- Request size doesn't increase
+- API structure doesn't change aside from the addition of a query parameter.
+- Keeps the translation logic out of runner
+Disadvantages:
+- Requires more logic in the API to choose the right translation - we already have some conversion logic, but we're currently trying to reduce the amount of logic in this repo.
```
theseanything, line 309: How would we implement storing the multiple locales in the form_documents in form-api?

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,336 @@
...
+Disadvantages:
+- More complexity in forms-admin. Probably needs API version increment.
+- Potential for overlap with API team's work.
+- Small runner change required to request the correct version.
+- Having two separate documents might risk the two versions going out of sync unexpectedly.
+- Would need to think about how we store the separate form documents - having them in one database column might be hard to query.
+
+## Decision
+
+Go with 'Selecting languages at request time' for now. Consider moving to 'Creating separate Welsh and English snapshots when making a form live' approach later if we add other languages and this this will be useful.
```
theseanything, line 329: Agree with "selecting languages at request time" - seems a sensible approach as requires minimal API schema changes and keeps response sizes the same. I wouldn't worry about making form-api do more work - that's a solvable problem, rather than a constraint. This also allows us to implement a non-breaking change to the API, i.e. just have it return English by default (e.g. when no locale parameter) specified.

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,336 @@
+# ADR044: Translate API response at request time
+
+Date: 2025-08-01
+
+## Status
+
+Accepted
+## Context
+
```
thomasiles, line 9: Do we need to reference https://github.com/alphagov/forms/blob/main/ADR/ADR007-store-form-text-data-as-objects.md

Which was an early, similar ADR which has remained at proposed.

Obviously things have moved on so much from then we could just dismiss it?

DavidBiddle: Good spot - I'd forgotten that ADR existed! Have updated to supersede that ADR, clarify that it was never implemented, and added a short note to this ADR about it as well.

ADR/ADR044-translate-api-response-at-request-time.md
DavidBiddle: My thinking was that we'd store it in a format analogous to the ones proposed in option one. 

I imagine storing them as side-by-side suffixed columns would be easiest:

```json
"name_en": "A form for testing file upload",
"name_cy": "Ffurflen ar gyfer profi uwchlwytho ffeiliau",
```

but I wasn't sure if this was worth including in the ADR or if we could leave it up to whoever picks up the implementation to choose an appropriate approach

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,417 @@
...
+}
+```
+
+These cases are essentially the same in how forms-runner deals with them.
+Benefits:
+
+- Few admin/api changes required
+  Disadvantages:
+- More runner logic, since we'd have to manage choosing the right translation and fallback behaviour in runner.
+- We'd get less benefit from the Mobility gem - we'd be using it for managing the columns and dealing with translations in admin, but we wouldn't be using its querying or fallback features in forms-runner.
```
stephencdaly, line 267: If the Mobility gem doesn't help us in forms-runner, how much benefit would using it with the suffixed columns offer over making the translatable columns have jsonb type? Dealing with the object per translatable field in forms-runner as opposed to the suffixed fields feels like it would be slightly easier and neater - but there might not be much difference.

Dealing with a jsonb type with translations per language feels like it would be fairly straightforward and not really warrant using a gem to help us, but are there particular benefits Mobility gives - for instance, does it make it easier to check that translations have been provided for all the translatable values in forms-admin?

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,417 @@
...
+- For forms with lots of content (e.g. multiple pages with a large amount of detailed guidance or what happens next text), this would increase the size of the request quite a lot since this information would be present in two different languages. Much of that content wouldn't be used for a given request, since the runner only shows one language at a time.
+- Additional languages would make it even more verbose.
+
+### 2. Selecting languages at request time
+
+In this approach, we would allow the runner to request a specific language when requesting the form from the API, and the API would only supply translations for the requested language.
+
+So `/forms/7/live?locale=en` would return:
```
lfdebrux, line 275: It's a little confusing that the URL parameter here is called `locale` while the attribute in the form document is called `language`; can we make the two consistent?

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,417 @@
...
+        "guidance_markdown": "Peth arweiniad",
+        "is_repeatable": false
+      },
+      "routing_conditions": []
+    }
+  ]
+}
+```
+
+Note that in this case we would still need to store the individual translations in the made_live_form JSON. We could use one of the formats described in option 1 for this.
```
lfdebrux, line 375: As we're dealing with the v2 API here, we should talk about form documents rather than made_live_forms

```suggestion
Note that in this case we would still need to store the individual translations in the form document JSON. We could use one of the formats described in option 1 for this.
```

ADR/ADR044-translate-api-response-at-request-time.md
lfdebrux: Hmm... I guess that makes sense for what's stored in the database, but for a form document response that has one translation in it I think it would be helpful for it to be tagged with the translation language; if only for debugging purposes!

ADR/ADR044-translate-api-response-at-request-time.md
lfdebrux: Does this ADR also supersede https://github.com/alphagov/forms/blob/main/ADR/ADR043-use-mobility-with-column-strategy.md then?

ADR/ADR044-translate-api-response-at-request-time.md
```diff
@@ -0,0 +1,411 @@
...
+- Few admin/api changes required
+  Disadvantages:
+- More runner logic, since we'd have to manage choosing the right translation and fallback behaviour in runner.
+- We'd get less benefit from the Mobility gem - we'd be using it for managing the columns and dealing with translations in admin, but we wouldn't be using its querying or fallback features in forms-runner.
+- For forms with lots of content (e.g. multiple pages with a large amount of detailed guidance or what happens next text), this would increase the size of the request quite a lot since this information would be present in two different languages. Much of that content wouldn't be used for a given request, since the runner only shows one language at a time.
+- Additional languages would make it even more verbose.
+
+### 2. Selecting languages at request time
+
+In this approach, we would allow the runner to request a specific language when requesting the form from the API, and the API would only supply translations for the requested language.
```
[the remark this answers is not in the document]
DavidBiddle: I think I'd expect it to default to the English version in that case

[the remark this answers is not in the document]
lfdebrux: Okay. So would there be any way to retrieve the actual form document stored in the database (with all the different translations side-by-side)?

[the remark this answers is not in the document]
DavidBiddle: Ah, admin will likely need that - so maybe that's what the queryless version should return. Runner should always have a locale set, so I don't think it should ever be requesting the queryless version, and defaulting to English probably doesn't provide any value.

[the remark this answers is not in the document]
DavidBiddle: I've incorporated this and the error response for unsupported languages into the text of the ADR, thanks!

ADR/ADR044-translate-api-response-at-request-time.md
DavidBiddle: Storing separate form documents for each language is essentially option 3. 

I've updated the ADR to recommend that option instead - I think it's a better fit given a) the concerns around essentially maintaining two separate schemas and b) plans to merge forms-api and forms-admin have become more concrete since I did the spike this ADR is based on

ADR/ADR044-translate-api-response-at-request-time.md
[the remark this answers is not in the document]
theseanything: Ah okay - feels like we could get away with doing multiple requests for debugging purposes. From the conversation above, translations are stored as separate documents anyways, we'd have to spend the extra engineering effort building a new response (for which I'm unclear the value).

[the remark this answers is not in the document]
lfdebrux: Yeah, my comment was based on the original proposal to store the translations side-by-side in one record.

theseanything (approved):

PR #225, opened 2025-08-19
Add ADR for removing forms-api app

lfdebrux:
# PR Checklist

---

# What

Add an Architecture Decision Record recording our decision to remove the forms-api app from our architecture.

# How to review

Rendered version: https://github.com/alphagov/forms/blob/ldeb-add-adr-remove-forms-api/ADR/ADRXXX-remove-forms-api.md

# Who can review

Any developers on GOV.UK Forms

### DIFF
--- a/ADR/ADR004-store-form-schemas-in-an-api.md
+++ b/ADR/ADR004-store-form-schemas-in-an-api.md
@@ -4,7 +4,7 @@ Date: 2022-02-18
 
 ## Status
 
-Accepted
+Superseded by [AD045: Remove forms-api](ADR045-use-forms-api.md)
 
 ## Context
 


ADR/ADRXXX-remove-forms-api.md
```diff
@@ -0,0 +1,45 @@
...
+- We will have fewer separate apps to maintain
+
+Negatives:
+- forms-admin will be in the critical path for submitting a form; the system will be less resilient
+- We won't have an external API for retrieving form documents
+
+This change is likely to be resource-neutral, in the best case it may use slightly less resources, but it is more likely it will increase our resource usage. Currently we have fewer forms-admin instances than forms-api instances running in production, but each forms-api instance uses more compute. It is difficult to predict what the right number of instances for the forms-admin app including API endpoints would be, we will probably over-provision initially and optimise resource usage later. In terms of costs we are currently on a [compute savings plan] that has pay a fixed annual cost for a pre-agreed amount of compute, so we may not see an increase of cost. In terms of CO2 usage our AWS systems are currently nominally carbon-neutral, as Amazon has "purchased renewable energy that compensated all or part of the emissions associated with [our] usage", and the change in this ADR is unlikely to increase our emissions beyond the point of being carbon-neutral.
```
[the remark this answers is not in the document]
lfdebrux: Yeah caching would help, I'd add a note to that effect. Caching would also help with versioning, which has been on our radar for a while as a potential issue (see https://trello.com/c/Oj9jMOhv/277-spike-exploration-avoid-impacting-users-when-a-live-form-is-changed, https://trello.com/c/FFJZq4n9/252-make-runner-request-specific-form-version-from-api), because currently our forms are not immutable.

ADR/ADRXXX-remove-forms-api.md
```diff
@@ -0,0 +1,45 @@
...
+However, maintaining multiple apps has resulted in increased toil. Having multiple services has also resulted in challenges around designing and maintaining the interface between forms-admin and forms-api.
+
+As the cost of compute is low at the moment, the cost of the additional developer time outweighs any savings due to efficient capacity management.
+
+We are currently refactoring forms-admin and forms-api to work with the v2 API described in [ADR034]. We have an opportunity now as part of this work to make a large architectural change, namely using the forms-admin app to serve the v2 API endpoints needed for forms-runner, and removing the need for forms-admin.
+
+[ADR004]: ./ADR004-store-form-schemas-in-an-api.md
+[ADR034]: ./ADR034-api-v2-for-forms.md
+
+## Decision
```
sarahseewhy, line 24: Some thoughts on migrating.

The decision states what the end state will be and I appreciate the step by step outline in this section. Will there be a phased migration (shadow traffic, canary release) or a big-bang switch? Is it worth calling that out specifically in the ADR?

What’s the rollback strategy if forms-admin can’t handle the load or the change introduces regressions? I very much doubt this will be the case but I want to ask the question just in case.

ADR/ADRXXX-remove-forms-api.md
```diff
@@ -0,0 +1,45 @@
...
+
+We will add endpoints to forms-admin that can be used to retrieve v2 API form documents in JSON format.
+
+We will change forms-runner to read form documents from these new endpoints in forms-admin.
+
+We will keep the new endpoints in forms-admin accessible from the private VPC only (keep them not accessible from the public internet), and not require authentication for their use.
```
sarahseewhy, line 30: The new endpoints will be VPC-only and unauthenticated. Is it worth validating this choice against a threat model to ensure it does not introduce unintended risk? 

I suspect doing this validation may be above and beyond what's necessary for this piece of work but in the spirit of "building with security in mind" I'm going to add it as a comment.

It might also be helpful to think about the pros/cons of a VPC-only and unauthenticated system. I'm in no way a security expert and the following pros/cons list is based on limited research and beginner understanding of security principles.

Pros:
- Simple and fast: no tokens/keys to manage, fewer moving parts.
- Small surface: no internet exposure, limited to private subnets and allowed security groups.
- Good first iteration: works as an outer fence while we add deeper controls.

Cons:
- Lateral‑movement risk: if the VPC is compromised an unauthenticated endpoint is especially at risk.
- No authentication == no caller identity?: if I understand the ADR correctly this would mean we couldn’t clearly answer “who called this?” -- might make forensics more difficult.

Like I said, this isn't my expertise but I want to capture some thoughts.

ADR/ADRXXX-remove-forms-api.md
```diff
@@ -0,0 +1,45 @@
+# ADRxxx: Remove forms-api
```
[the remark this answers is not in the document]
lfdebrux: Ah, I was going to amend the commit when it's closer to the time that this can be merged, as I'm expecting some other ADRs including the Welsh one to be merged first.

ADR/ADRXXX-remove-forms-api.md
```diff
@@ -0,0 +1,45 @@
...
+
+We will retire the forms-api app and remove it, its database, and associated systems, networks and configuration files.
+
+## Consequences
+
+Positives:
+- We will have fewer separate apps to maintain
+
+Negatives:
+- forms-admin will be in the critical path for submitting a form; the system will be less resilient
```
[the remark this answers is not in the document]
lfdebrux: Thanks for your thoughts!

I'm happy to expand on this statement a small amount in the document, but I think the discussions here are also a good record of how we came to a decision.

I guess I'd start by asking whether you agree that the system would be less resiliant?

My rationale for saying so was that with forms-api separate from forms-admin, that means forms-admin can go down without form submissions being affected; so I think all other things being equal, the system will be less resilient without forms-api. I'm happy to hear counter-arguments though!

Of course, that is with all other things being equal, so we could think about mitigations, but I think we would probably first need to determine whether the loss in resilience is worth the extra engineering effort, and that's not really the problem we're trying to address in this ADR. So I'd probably just leave it at saying "we could investigate other ways to improve resilience".

ADR/ADRXXX-remove-forms-api.md
lfdebrux: Also not my area, I think though at the moment the plan is for the API to be limited to a handful of read-only endpoints, which limits the damage that could be done with the unprotected API to exfiltration and reputational damage. Basically I think if the VPC was compromised there would be juicier targets.

No caller identity would be a little annoying, although could be mitigated somewhat by looking at headers including the request host and `X-Amzn-Trace-Id`; we have logging for those in forms-api in https://github.com/alphagov/forms-api/blob/main/app/controllers/application_controller.rb#L23-L31 that we could copy across to forms-admin if there isn't already similar there.

theseanything: Yeah, risk here is low - endpoints are readonly and expose data thats likely publicly available anyway (we don't authenticate draft forms from runner). This shouldn't be a blocker for this work, and doesn't rule out us improving the security posture on these internal endpoint at a later date.

sarahseewhy: This is great, thank you both for your answers! This was another due diligence question, and concur with Sean it feels like the risk is low given the context.

theseanything (dismissed):

DavidBiddle (approved):

PR #226, opened 2025-09-05
Add ADR for using Mobility's table backend

DavidBiddle:
# What

We previously made an ADR proposing using the Mobility gem with its `:column` backend (https://github.com/alphagov/forms/pull/213). 

This PR adds a new ADR changing our approach to use a different Mobility backend (:table) - we think this will reduce the amount of effort overall.

# How to review

# Who can review

Any dev other than @DavidBiddle

### DIFF
--- a/ADR/ADR043-use-mobility-with-column-strategy.md
+++ b/ADR/ADR043-use-mobility-with-column-strategy.md
@@ -4,7 +4,7 @@ Date: 2025-06-09
 
 ## Status
 
-Accepted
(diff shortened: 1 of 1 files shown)


lfdebrux:
Does this mean we'll need to do a data migration for moving existing content into the translations table?

DavidBiddle:
@lfdebrux I don't think a data migration for our existing content is strictly required, but it would make our lives easier in the long run

PR #248, opened 2026-03-24
Add ADR048: Immutable Form Versioning

theseanything:
This ADR proposes changing GOV.UK Forms so each published form becomes a permanent, numbered version.

The main point is: once a form is published, that exact version will never change. If the form owner edits and republishes it, the system creates version 2, then 3, and so on. Users who started version 1 can continue with version 1, while new users get the latest live version.

Key benefits:

**Better user journeys:** people filling in a form are not disrupted if the form is changed or archived mid-way.
**Support for future form versioning in Forms admin:** form creators could see, manage, and understand the history of published versions.
**Support future implementation of Save and Return:** users could return to the same version of the form they originally started.
**Clearer submission history:** every submission records which form version it used.
**Better downstream processing:** processors no longer need to guess which form version (and it's structure) a submission relates to.
**Improved performance, reliability and resilience to traffic spikes:** published versions can be cached because they never change. If Forms Admin breaks, it's less likely to affect forms runner.
**Stronger audit trail:** the full history of published form versions is preserved.

The main product tradeoff is around archiving. Today, archiving can effectively stop people from submitting or progressing with the form. With immutable versions, archiving would stop new users from starting, but users already on a version may still be able to finish. For forms with strict legal or policy deadlines, we would need to re-implement that functionality explicitly.

### DIFF
--- a/ADR/ADR048-immutable-form-versioning.md
+++ b/ADR/ADR048-immutable-form-versioning.md
@@ -1,0 +1,63 @@
+# ADR048: Immutable Form Versioning
+
+Date: 2026-03-24
+
+## Status
+
+Accepted
+
+## Context
+
+GOV.UK Forms currently uses a mutable document-based system for managing form lifecycle states. The `Form` model has an associated `FormDocument` model, with documents tagged as `draft`, `live`, or `archived`. The current API (v2) exposes these via:
+
+- `GET /api/v2/forms/:form_id/draft`
+- `GET /api/v2/forms/:form_id/live`
+- `GET /api/v2/forms/:form_id/archived`
+
+The `FormStateMachine` manages transitions between states (`draft`, `live`, `live_with_draft`, `archived`, `archived_with_draft`), and the `FormDocumentSyncService` synchronises the JSON content into `FormDocument` records when these transitions occur.
+
+This approach has several limitations:
+
+1. The `/live` endpoint is mutable. The content behind `/api/v2/forms/:form_id/live` changes each time a form is re-published, so consumers must always re-fetch. This prevents effective caching.
+2. No explicit link between a submission and the form version it was submitted against. When a form is updated and re-published, there is no reliable way to identify which version of the form a given submission relates to. This makes it difficult to group submissions by form version or detect when a form changed between batch submission deliveries.
+3. Mid-journey disruption. If a form creator publishes a new version or archives a form while a user is part-way through filling it in, the form can change or disappear mid-journey. There is no mechanism for in-progress users to continue with the version they started.
+
+## Decision
+
+We will introduce an immutable versioning model for published forms, exposed through a new v3 API. The key changes are:
+
+### New API endpoints
+
+
+| Endpoint                                            | Description                                                                                              |
+| --------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
+| `GET /api/v3/forms/:form_id/draft`                  | Returns the current draft form document JSON (mutable, changes as the form creator edits)                |
+| `GET /api/v3/forms/:form_id/versions/:form_version` | Returns an immutable, versioned form document. Once created, this content never changes.                 |
+| `GET /api/v3/forms/:form_id/latest`                 | Returns the latest live version of the form (a redirect or alias to the most recently published version) |
+
+
+### Lifecycle
+
+- Draft state: A form being edited has a draft available at `/api/v3/forms/:form_id/draft`. This behaves similarly to today.
+- Publishing (making live): When a form is made live, a new immutable version is created and assigned an incrementing version identifier (e.g. `1`, `2`, `3`). It becomes available at `/api/v3/forms/:form_id/versions/:form_version`. The `/api/v3/forms/:form_id/latest` endpoint points to this new version.
+- Archiving: When a form is archived, `/api/v3/forms/:form_id/latest` and `/api/v3/forms/:form_id/draft` return `404` (or `410 Gone`). However, all previously published versions remain available at `/api/v3/forms/:form_id/versions/:form_version` because they are immutable.
+
+### Content removal
+
+Immutability prevents deletion as part of normal operations. A separate process will be needed for exceptional cases where published content genuinely must be removed (e.g. GDPR erasure).
+
+## Consequences
+
+### Positive
+
+- Cacheable published forms. Versioned form documents at `/api/v3/forms/:form_id/versions/:form_version` can be cached indefinitely by consumers (e.g. forms-runner, CDNs), significantly reducing load on the API and improving latency for form rendering.
+- Submissions linked to form versions. Each submission can explicitly reference the `form_version` it was submitted against. This enables grouping submissions by version and helps people processing submissions to know exactly which questions were asked.
+- Graceful publishing. Users who have already started filling in a form can continue submitting against the version they began with, even if the form creator publishes a new version in the meantime.
+- Graceful archiving. When a form is archived, new users can be prevented from starting the form while users who have already started can finish and submit against the version they are on.
+- Reverting to previous versions. Preserving all published versions makes it easier to implement future features allowing form creators to revert to a previous version of a form.
+- Audit trail. The full history of published form versions is preserved and addressable.
+
+### Negative
+
+- Migration and data model complexity. Existing consumers (primarily forms-runner) will need to be updated to use the v3 API, requiring a transition period running both APIs in parallel. The `FormDocument` model (or a new model) will need to support version numbering alongside or in place of the current tag-based system (`draft`, `live`, `archived`).
+
--- a/ADR/ADR048-immutable-form-versioning.md
+++ b/ADR/ADR048-immutable-form-versioning.md
@@ -42,6 +42,12 @@ We will introduce an immutable versioning model for published forms, exposed thr
 - Publishing (making live): When a form is made live, a new immutable version is created and assigned an incrementing version identifier (e.g. `1`, `2`, `3`). It becomes available at `/api/v3/forms/:form_id/versions/:form_version`. The `/api/v3/forms/:form_id/latest` endpoint points to this new version.
 - Archiving: When a form is archived, `/api/v3/forms/:form_id/latest` and `/api/v3/forms/:form_id/draft` return `404` (or `410 Gone`). However, all previously published versions remain available at `/api/v3/forms/:form_id/versions/:form_version` because they are immutable.
 
+### Hard submission deadlines
+
+For legal or policy reasons, some forms may need a strict cutoff time after which no new submissions are permitted. Archiving would no longer act as a way to cut off in-progress journeys and prevent any future submissions.
+
+However, this behaviour can be re-implemented. For example, this could be a deadline timestamp attribute on the form that `forms-runner` checks before displaying the form or accepting a submission. This is more explicit and reliable than using archiving as a proxy for a hard stop. It would also allow form owners to schedule a cutoff in advance.
+
 ### Content removal
 
 Immutability prevents deletion as part of normal operations. A separate process will be needed for exceptional cases where published content genuinely must be removed (e.g. GDPR erasure).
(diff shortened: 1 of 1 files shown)


ADR/ADR048-immutable-form-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+
+We will introduce an immutable versioning model for published forms, exposed through a new v3 API. The key changes are:
+
+### New API endpoints
+
+
+| Endpoint                                            | Description                                                                                              |
+| --------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
+| `GET /api/v3/forms/:form_id/draft`                  | Returns the current draft form document JSON (mutable, changes as the form creator edits)                |
+| `GET /api/v3/forms/:form_id/versions/:form_version` | Returns an immutable, versioned form document. Once created, this content never changes.                 |
```
stephencdaly, line 34: Do we want live form documents to be actually immutable - or immutable from an end user perspective? The main issue we might encounter here is when we make internal changes to form structure such as adding/removing/changing attributes.
It would make it harder to make changes to forms-runner and hard to support users reverting changes if we don't also update the structure of historic form documents when we make breaking changes to the format.

There might be some things we can do to lessen the need for altering existing form documents. Currently when we add attributes to a form, we always back-fill the attribute on form documents. Potentially we can change our approach to this and fall back on default values when an attribute isn't present rather than updating it in the database.

theseanything: V good point — I think ultimately we want the semantic content to be immutable.

We could handle changes in schema in a few ways:

1. Include a schema version in the form document. This makes it explicit to consumers how to interpret the content. It may mean we'd need to support multiple schema versions, but in practice, making sure most changes are backwards-compatible minimises the complexity around this. (Could reserve breaking schema changes for a new API version.)

2. Weaken the immutability by keeping the semantic content the same but allowing the schema to be updated. i.e. migrate old documents to newer schemas. We wouldn't be able to cache "forever", but schema changes are probably infrequent enough that caching would still be useful.

theseanything: I'd probably lean towards 1. as it gives stronger guarantees for the API, in way that is less likely to break forms-runner.

ADR/ADR048-immutable-form-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Cacheable published forms. Versioned form documents at `/api/v3/forms/:form_id/versions/:form_version` can be cached indefinitely by consumers (e.g. forms-runner, CDNs), significantly reducing load on the API and improving latency for form rendering.
+- Submissions linked to form versions. Each submission can explicitly reference the `form_version` it was submitted against. This enables grouping submissions by version and helps people processing submissions to know exactly which questions were asked.
+- Graceful publishing. Users who have already started filling in a form can continue submitting against the version they began with, even if the form creator publishes a new version in the meantime.
+- Graceful archiving. When a form is archived, new users can be prevented from starting the form while users who have already started can finish and submit against the version they are on.
+- Reverting to previous versions. Preserving all published versions makes it easier to implement future features allowing form creators to revert to a previous version of a form.
+- Audit trail. The full history of published form versions is preserved and addressable.
+
+### Negative
+
```
lfdebrux, line 83: Another negative is increased complexity for form processors; making it so that an old version of a form can be submitted after the new version has been made live means that submission processors need to be able to handle both old and new version submissions for a time.

ADR/ADR048-immutable-form-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+This approach has several limitations:
+
+1. The `/live` endpoint is mutable. The content behind `/api/v2/forms/:form_id/live` changes each time a form is re-published, so consumers must always re-fetch. This prevents effective caching.
+2. No explicit link between a submission and the form version it was submitted against. When a form is updated and re-published, there is no reliable way to identify which version of the form a given submission relates to. This makes it difficult to group submissions by form version or detect when a form changed between batch submission deliveries.
+3. Mid-journey disruption. If a form creator publishes a new version or archives a form while a user is part-way through filling it in, the form can change or disappear mid-journey. There is no mechanism for in-progress users to continue with the version they started.
+
+## Decision
+
+We will introduce an immutable versioning model for published forms, exposed through a new v3 API. The key changes are:
```
lfdebrux, line 27: I think there's a way we could achieve the same aims and outcomes of this ADR but without needing a new API version; would you be interested in that or would you prefer to have a new API version?

ADR/ADR048-immutable-form-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+Immutability prevents deletion as part of normal operations. A separate process will be needed for exceptional cases where published content genuinely must be removed (e.g. GDPR erasure).
+
+## Consequences
+
+### Positive
+
+- Cacheable published forms. Versioned form documents at `/api/v3/forms/:form_id/versions/:form_version` can be cached indefinitely by consumers (e.g. forms-runner, CDNs), significantly reducing load on the API and improving latency for form rendering.
+- Submissions linked to form versions. Each submission can explicitly reference the `form_version` it was submitted against. This enables grouping submissions by version and helps people processing submissions to know exactly which questions were asked.
+- Graceful publishing. Users who have already started filling in a form can continue submitting against the version they began with, even if the form creator publishes a new version in the meantime.
+- Graceful archiving. When a form is archived, new users can be prevented from starting the form while users who have already started can finish and submit against the version they are on.
```
lfdebrux, line 76: For some forms, it might be a legal requirement that submissions cannot be made after a certain time; in the past I think we've had it where form publishers have used the archive this form feature to enforce that. Changing it so form fillers can submit the form after the form is archived might make that more difficult?

ADR/ADR048-immutable-form-versioning.md
theseanything: That's true, but I think form processors already have this problem today. Unless a form owner can update the form and their processing system at exactly the same time, there's always a window where submissions come in against a version the processor wasn't expecting. The current setup just makes that invisible.

Explicit version numbers actually make it easier to handle. A processor can see which version a submission was made against and branch accordingly (e.g. if version 2, do X, otherwise do Y), rather than having to infer what changed from the shape of the data.

ADR/ADR048-immutable-form-versioning.md
theseanything: Ooo! Yes always interested in hearing ideas!

ADR/ADR048-immutable-form-versioning.md
theseanything: Yup, fair point. I think the number of forms requiring a hard legal deadline is probably much smaller than the number of form fillers who'd benefit from not being kicked out mid-form (where the deadline is usually softer).

For those edge cases, we could reintroduce a hard stop as a form attribute. Something like a deadline timestamp that forms-runner checks before displaying the form or accepting a submission. That would actually be more explicit and reliable than relying on archiving to enforce a legal cutoff, and it'd have the added benefit of letting form owners schedule the cut-off in advance rather than having to manually archive at the right moment.

ADR/ADR048-immutable-form-versioning.md
lfdebrux: Okay, so as part of the changes for this ADR we'll expose the version information to form processors, and inform them of the change?

ADR/ADR048-immutable-form-versioning.md
lfdebrux: I think if we added the `/versions` endpoints you're suggesting to the v2 API, and added a `version` link to the form documents, that would cover the same functionality more or less?

So for a form document that's been made live, regardless of where it's retrieved from, you'd see something like:

```
{
  form_id: "aB123x",
  name: "Test form",
  ...
  links : {
    version: "https://api.forms.service.gov.uk/api/v2/forms/aB123x/versions/9"
  }
}
```

ADR/ADR048-immutable-form-versioning.md
theseanything: Add section to ADR about the need to re-implement this functionality.

ADR/ADR048-immutable-form-versioning.md
theseanything: Yup - added info that submissions would reference the form version. And yup, would be good to inform them.

ADR/ADR048-immutable-form-versioning.md
theseanything: Ah thats a nice idea. 

I think my main concern with doing this in v2: we’d end up with a slightly confusing API shape where the old state-based endpoints still exist alongside the new version-based ones.

For example, it becomes unclear what /api/v2/forms/:id/archived is meant to represent

ADR/ADR048-immutable-form-versioning.md
theseanything: I've updated the ADR with a section on API changes.

ADR/ADR048-immutable-form-versioning.md
lfdebrux: I think it still represents the form document as it was when it was archived? We have that so that it's easy for forms-runner to preview the form after its been archived but changes have been made to it. With endpoints for versions we could alternatively do that by sending them to a preview for that version, which I guess what you're proposing currently, but both still seem valid to me?

It doesn't help that I've thought about this so much that it's hard to imagine looking at it for the first time, so I'm unsure about what you mean about it being unclear 😅 

When I was thinking about the v2 API I thought about the state endpoints as branches/tags in a git repository, or symbolic links in a filesystem, so that's where I'm coming from.

theseanything: Yeah, I agree with the referential model. Thinking of the state endpoints as branches/tags/symlinks makes sense to me.

This ADR is kind of an iteration of that in which we would start exposing the “commits” directly, i.e. immutable form versions. I think `/live` and `/archived` probably shouldn’t really be document endpoints any more; they should just be redirects (e.g. symbolic links, pointers) to a specific immutable version.

Where I think this starts to get awkward in v2 is that the existing endpoints currently return embedded form documents in their response. That feels more like a contract change than an additive one, which is why I’m leaning towards a new API version. Otherwise v2 ends up with two overlapping models.

Agree with you on `/archived`. We probably do need to keep it, at least for compatibility, so we don’t break the existing forms-runner `preview-archived` behaviour.

So I don’t think your suggestion is invalid; adding version links to v2 could probably cover much of the functionality. My concern is more that we'd be stuck with /live and /archived returning the form document in the response unused - which would be confusing.

ADR/ADR048-immutable-form-versioning.md
stephencdaly: Where is the bit in the ADR that talks about how we'll handle schema changes? Sorry I might be missing it.

ADR/ADR048-immutable-form-versioning.md
```diff
@@ -0,0 +1,83 @@
...
+We will introduce an immutable versioning model for published forms, exposed through a new v3 API. The key changes are:
+
+### New API endpoints
+
+
+| Endpoint                                            | Description                                                                                              |
+| --------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
+| `GET /api/v3/forms/:form_id/draft`                  | Returns the current draft form document JSON (mutable, changes as the form creator edits)                |
+| `GET /api/v3/forms/:form_id/versions/:form_version` | Returns an immutable, versioned form document. Once created, this content never changes.                 |
+| `GET /api/v3/forms/:form_id/latest`                 | Returns the latest live version of the form (a redirect or alias to the most recently published version) |
```
lfdebrux: If we're going with a new v3 API:

I find it a bit confusing that `:form_id/draft` returns a form document but `:form_id/latest` returns a link. Could we make the endpoint to get the latest form document a bit different?

What about `:form_id/versions/latest`? Or `:form_id/tags/latest`?

theseanything: Or what about the inverse? `:form_id/versions/draft` being a special form version which is mutable. That way everything under :form_id/versions returns a form document?

lfdebrux: Yeah that works too!

stephencdaly (approved):

PR #252, opened 2026-04-15
Add Welsh translation state diagram

thomasiles:
## Add a simple explanation of how Welsh translations interact with the database fields and tasklist.

Trello: https://trello.com/c/6ED8Pbcc/2895-create-a-sequence-diagram-for-adding-welsh-translation

### DIFF
--- a/diagrams/state-diagrams/welsh-translation-states-and-fields.md
+++ b/diagrams/state-diagrams/welsh-translation-states-and-fields.md
@@ -29,8 +29,8 @@ stateDiagram-v2
 
 ## available_languages
 
-`available_languages`, is an array of language codes. It's `[:en]` by default, which signals that the form is only available in English.
-Forms with a Welsh translation have `[:en, :cy]` in the array.
+`available_languages`, is an array of language codes. It's `["en"]` by default, which signals that the form is only available in English.
+Forms with a Welsh translation have `["en", "cy"]` in the array.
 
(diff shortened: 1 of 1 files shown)


diagrams/state-diagrams/welsh-translation-states-and-fields.md
```diff
@@ -0,0 +1,53 @@
...
+    In_Progress --> Complete:Enter Welsh, save unfinished
+    Complete --> In_Progress:Invalidate Welsh
+```
+
+## New fields on Form
+
+## available_languages
+
+`available_languages`, is an array of language codes. It's `[:en]` by default, which signals that the form is only available in English.
+Forms with a Welsh translation have `[:en, :cy]` in the array.
```
DavidBiddle: I think we use strings rather than symbols in the array:
```suggestion
`available_languages`, is an array of language codes. It's `["en"]` by default, which signals that the form is only available in English.
Forms with a Welsh translation have `["en", "cy"]` in the array.
```

DavidBiddle (commented):
Have raised one comment but it looks good overall - feels like the right level of detail, clearly written, diagram makes sense to me 🎉

diagrams/state-diagrams/welsh-translation-states-and-fields.md
thomasiles: Good point, thanks!

DavidBiddle (approved):

PR #253, opened 2026-04-16
Add ADR049: Remove user research environment

theseanything:


chao-xian (approved):

DavidBiddle (approved):
The decision makes sense to me and the ADR explains the case for it clearly 🎉

cadmiumcat (approved):
This seems like the right call 🙌 

I was a bit worried about dev not being able to meet all the needs for whatever user research we're carrying out. But I think it'd probably be easier to change the application side of things (e.g. allow users to sign up with gov.uk email addresses) than to recreate a user-research environment.

I was also worried about having to recreate the user-research environment. But it sounds like this is unlikely. If we do ever have to, the infra team will need enough notice to figure this out.

PR #263, opened 2026-09-08
Add ADR052 on storing analytics

theseanything:
We want to measure submissions over time and the number of live forms, keeping the data for at least two years. 

Existing tools (OTel and Classic CloudWatch metrics, GA4, Splunk, Logs Insights, the runner database) fall short on accuracy, cost, cardinality, retention or ownership.

This record sets out the two credible options, S3 Tables queried with Athena or a separate Aurora cluster, for the team to choose between. It is marked Proposed until that choice is made.

### DIFF
--- a/ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
+++ b/ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
@@ -1,0 +1,107 @@
+# ADR052: Store analytics in S3 Tables and query with Athena, or in a separate Aurora database
+
+Date: 2026-09-07
+
+## Status
+
+Proposed
+
+This record sets out two options for team discussion rather than a single decision. Once a choice is made it will be recorded here and the status changed to Accepted.
+
+## Context
+
+We want to measure key performance indicators for GOV.UK Forms, starting with:
+
+- the number of submissions over time
+- the number of live and archived forms over time
+
+We need at least 2 years of data for year-on-year comparisons. Longer retention is preferred provided it is cheap and efficient for us.
+
+These are business analytics rather than operational monitoring. Accuracy matters. The data does not need to be real-time, but having up-to-date figures is useful.
+
+forms-runner already logs a structured `form_submission` event for every submission.
+
+Recording the number of submissions per form and the number of forms per organisation is not a necessity, but the granularity would be useful if we later wanted to reuse the metrics for form-level analytics.
+
+### Options considered and rejected
+
+**CloudWatch OpenTelemetry metrics (tried).** This is AWS's current, recommended metrics model, billed per GB ingested with 15 months of retention. Metrics are statistical aggregates, not a record of events, so should not be used for anything that needs exact figures. For example, counters are aggregated in-process and exported on an interval, so counts are lost if a task exits before export, and PromQL `rate()` and `increase()` extrapolate across the query window and can return non-integer results. The OpenTelemetry SDK also caps each metric at 2,000 attribute combinations by default, after which per-form measurements collapse into a single overflow bucket.
+
+**CloudWatch Classic metrics.** Every unique dimension combination is a separately billed metric at $0.30 per metric per month, so a per-form dimension costs roughly $0.30 per form per environment per month and grows with every form we host. Metrics expire after 15 months.
+
+**Google Analytics 4.** Client-side events only fire after a user accepts usage cookies, so they undercount. The server-side Measurement Protocol requires a user id on every event, so we would have to create pseudo ids. Sending submissions with fabricated ids either inflates user counts or lands them under "(not set)", polluting our real user metrics. A per-form dimension exceeds GA4's 500-value high-cardinality threshold, so less common forms are condensed into an "(other)" row. Event-level data is retained for at most 14 months on a standard property.
+
+**Splunk.** We keep logs for 12 months, it is not a tool we own, and we intend to move away from it.
+
+**CloudWatch Logs Insights over existing logs.** This needs no new infrastructure, but every query scans the whole runner log group rather than just submission events, so we pay for and wait on data we do not need. Keeping the whole log group for years to preserve a handful of events would be wasteful.
+
+**The forms-runner database.** This would grant observability tooling access to a database with potentially sensitive information, and analytics queries could affect production performance.
+
+**Amazon Redshift Serverless or OpenSearch.** Both are capable, but their cost far exceeds what a few small tables need.
+
+## Decision
+
+Two options are outlined below. The following applies to both:
+
+- Only non-sensitive fields are stored: form identifiers and metadata, never answers or personal data.
+- Isolated from production databases, so analytics queries cannot affect production performance.
+- We can set our own retention limits.
+- Both can be queried by data visualisation tools (each has a Grafana datasource).
+- Data can be moved between either option if we need to change decision.
+
+### Option A: Amazon S3 Tables, queried with Amazon Athena
+
+Amazon S3 Tables is storage optimised for analytics workloads. Tables are stored in Apache Iceberg format and can be queried by any engine that supports Iceberg. Athena is a serverless query engine that supports Iceberg and S3 Tables.
+
+Each environment gets an S3 table bucket with a `forms` namespace.
+
+Initially, events are taken from the structured log lines the applications already emit: a CloudWatch Logs subscription filter selects them, a small Lambda transform maps them to the table schema, and Kinesis Data Firehose writes them to the table. This needs no application changes. Later, the applications could write events directly, removing the dependency on the log format. The decision here is primarily where the data lives and how it is queried.
+
+Pros:
+
+- Negligible cost, less than $10 per year per environment: the main costs are S3 storage and Athena's per-query charge.
+- Fully managed. S3 Tables handles compaction and snapshot maintenance; Firehose, Lambda and Athena have no servers.
+- S3 durability and availability.
+- Fast to set up, already have working prototype, no application changes.
+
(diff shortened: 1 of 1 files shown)


ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
```diff
@@ -0,0 +1,107 @@
...
+- Fast to set up, already have working prototype, no application changes.
+
+Cons:
+
+- Unfamiliar services: S3 Tables, Firehose and Athena.
+- Firehose delivery is at-least-once, so queries may need de-duplication or the transform needs an idempotency key.
+- Table schemas are defined in Terraform, but the provider does not support partitioning yet (not a big concern at our scale).
+- Athena queries take seconds, too slow for the request path. Showing per-form statistics in forms-admin would need a faster way to query.
+
+### Option B: a separate Aurora PostgreSQL cluster for analytics
```
theseanything, line 74: Think this is my preferred option atm - as whilst it's more expensive, it's a comparatively small infra cost. We are unlikely to run into performance issues at our scale and is more flexible.

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
```diff
@@ -0,0 +1,107 @@
...
+- Fully managed. S3 Tables handles compaction and snapshot maintenance; Firehose, Lambda and Athena have no servers.
+- S3 durability and availability.
+- Fast to set up, already have working prototype, no application changes.
+
+Cons:
+
+- Unfamiliar services: S3 Tables, Firehose and Athena.
+- Firehose delivery is at-least-once, so queries may need de-duplication or the transform needs an idempotency key.
+- Table schemas are defined in Terraform, but the provider does not support partitioning yet (not a big concern at our scale).
+- Athena queries take seconds, too slow for the request path. Showing per-form statistics in forms-admin would need a faster way to query.
```
stephencdaly, line 72: There's a reasonable demand for funnel analytics from form creators, and we've been talking about adding metrics for save and return.

I think it's a reasonable assumption that certain metrics/analytics we make available to form creators we'll also want to use for internal performance analytics.

So I think a solution that works for both purposes seems best, otherwise we have to maintain emitting metrics from forms-runner in 2 different ways.

cadmiumcat: I agree. 

I think it's worth adding our requirements for these to the `context`. In the context of KPIs this last point wouldn't be a so much of a con.

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
cadmiumcat: Do we have a sense of how other teams do it?

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
theseanything: I've updated the context - made the proposal for using a separate Aurora DB

ADR/ADR052-store-analytics-in-aurora.md
```diff
@@ -0,0 +1,71 @@
...
+We will store analytics in a new Aurora PostgreSQL cluster, separate from the forms-admin and forms-runner databases, holding only analytics tables. Each environment gets its own cluster.
+
+The applications write events to it directly through a second Rails database connection, rather than deriving events from logs. Only non-sensitive fields are stored: form identifiers and metadata, never answers or personal data.
+
+We chose Aurora because:
+
+- It is familiar. We already run Aurora PostgreSQL, so there are no new services to learn and we can reuse existing Terraform, backup and access patterns.
+- Writes are transactional and exactly-once, with no log parsing or de-duplication, which suits data where accuracy matters.
+- Indexed rows give low-latency queries, so the same store can later serve per-form analytics inside forms-admin.
+- It is isolated from production databases, so analytics queries cannot affect production performance.
```
cadmiumcat, line 60: Do we know how big of a problem this actually is?

I think we have relatively low traffic. I'm wondering whether we could make it work with the existing clusters. Is it worth considering adding read replicas to deal with any performance concerns? I believe this is cheaper than adding a new cluster

theseanything: So I think - you still pay the same ACU costs, but uses the existing storage (tho that is negligible anyways) - however, we might be able to allow the read replica scale down to zero (but the long start up times, wouldn't be suitable prod)?

I guess the other thing is how we limit data access, because it'd be the forms-runner db with the submission data - which would require creating another (Postgres) Role in the DB. You could do it as a Rails migration? but kinda messy, you need to also get the password into Secrets Manager - and co-ordinate when the sequence of events. 

Not convinced the complexity of keeping the data secure is worth the cost savings.

cadmiumcat: I had assumed we'd create a separate database within the same cluster. So that the isolation is provide at that level. I think the access is managed per database, rather than per cluster
But I might be getting confused about how that works 🤔

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
theseanything: GOV.UK Notify, store their notifications sent in a db. Haven't heard from Pay or anyone else

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
theseanything: I'm still very uncertain, not totally certain an Aurora cluster is the right call.

Our actual requirement is so small: submission events and the number of forms over time, for an internal dashboard. Future analytics like funnels are still unclear and I have doubts our own database is the right home for them anyway, since they need to track users across interactions. That's better served by GA or a purpose-built product analytics tool.

Against that, an always-on cluster per environment, plus upgrades, backups and credentials, feels like a lot of cost (it'll be like $1000 - $1500 all up vs less than $50) + maintenance. The S3 Tables prototype is essentially free, covers the use case we understand today, and would be much quicker in getting us going with our KPIs. 

Now I'm leaning towards starting with S3 Tables and revisiting if we outgrow it. Keen to hear if anyone sees it differently.

cadmiumcat: I think there's a distinction between adding a new Aurora cluster and adding a new database to existing clusters. The cost of the latter is much lower.

From the proposal I assume you meant a cluster, but have we considered adding a database to an existing cluster?

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
cadmiumcat: Thank you. 
Do we know if they have any issues/wish they had done it differently? Have they struggled with performance

ADR/ADR052-store-analytics-in-aurora.md
theseanything: So you can create another logical "database" in postgres cluster - but you still need to manage permissions to that inside postgres (not something you can do with IAM). Within the physical cluster - you can create a "read replica", but that accesses the same underlying data (with all the logical databases).

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
theseanything: Possibly, considering the outlined use case is our own internal KPIs and at such a small scale - I'd like to take the risk and get something implemented sooner, rather than waiting to hear back from them.

ADR/ADR052-store-analytics-in-s3-tables-or-aurora.md
theseanything: Yes I did mean cluster - (tho I've interchangeable used Aurora DB). I think what you're suggesting is have we considered the cost of writing the metrics to the existing cluster (logically separated in someway).

From the previous comment thread - my concern is the complexity around managing access to the sensitive production data. I.e. you want to Grafana to only have read access to the metric tables and nothing else.

If I've misunderstood, or maybe its easier than I realise - might be helpful if you could describe to how exactly you would set that up?

ADR/ADR052-store-analytics-in-aurora.md
stephencdaly: Could we definitely not go with the option of storing the metrics in the admin db, using that data to drive metrics we provide to form creators (instead of cloudwatch), and offer a simple dashboard and download of these stats for internal purposes?

theseanything: We could, but I think it's quite a lot more complex and effort for right now.

E.g. building and maintaining a metrics API on forms-admin, new tables, and either dashboards in admin or a Grafana integration on top. Grafana is happening anyway for observability, so we'd end up with two places to look.

It also tightens the coupling between runner and admin - when we're looking to soften the dependency (e.g. caching the form documents).

I think the cost of switching to this later is cheap enough once we better know what analytics we want to provide to users. (But at the moment that functionality is not priority).

stephencdaly: I would argue that it's simpler for right now. You're right that it does tighten the coupling between runner and admin, though. But if you think option in the ADR is simpler I'm happy to be overruled on that.

theseanything: I think basically because the work for S3 Tables/Athena is already done (in the prototype). Will go for that now - and we can easily rip (as it doesn't require any application changes), and can pivot to forms-admin storing them  if thats right decision when we look at form analytics for users.

PR #264, opened 2026-09-08
Add ADR053: Use Grafana

theseanything:
Adds ADR053 for using Grafana as our tool for dashboards and data visualisation.

### DIFF
--- a/ADR/ADR053-use-grafana.md
+++ b/ADR/ADR053-use-grafana.md
@@ -1,0 +1,49 @@
+# ADR053: Use Grafana
+
+Date: 2026-09-08
+
+## Status
+
+Accepted
+
+## Context
+
+GOV.UK Forms has no central tool for visualising data. We want to show our key performance indicators (KPIs), which will likely be stored in a SQL database, to the whole team.
+
+We currently have 2 options for dashboards. Both have limitations:
+
+1. Splunk is limited to the logs stored in Splunk. We do not own it, and we want to move away from it so we can centralise our observability stack.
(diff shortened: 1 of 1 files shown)


ADR/ADR053-use-grafana.md
```diff
@@ -0,0 +1,49 @@
...
+
+Grafana will be available to the whole team, not only engineers.
+
+### Implementation options
+
+This decision does not cover how we implement Grafana. However, there are 2 options for reference:
+
+- Amazon Managed Grafana (AMG) is a fully managed solution. It is currently limited to Grafana version 12, while the latest release is version 13. Sign-in is only available through SAML or IAM Identity Center. EE have not set up Entra or Identity Center yet, so we cannot use AMG today. It is priced per user and would cost roughly $1,000 a year for the production environment (for example, 3 editors and 10 viewers).
+- Self-hosting open source Grafana on ECS would need an ECS Fargate task, an Application Load Balancer (ALB) and an Aurora PostgreSQL database for Grafana's own state. We can run the latest version of Grafana and use GitHub sign-in for now, as GOV.UK and GOV.UK Notify do. GOV.UK Pay also self-hosts. It would cost roughly $1,000 a year for the production environment.
```
cadmiumcat, line 33: I understood from our the tech weekly catch up that self-hosting Grafana was not a good option for us at all (although we didn't go into the details of it). Has this changed?

theseanything: Sorry I can't remember what was said - I this doc reflects my most current views. But in terms of this ADR, we don't need to agree on the implementation.

stephencdaly (approved):

PR #266, opened 2026-09-17
Add ADR054: Store saved answers in the forms-runner database

stephencdaly:
ADR for the "Save & Return" feature that proposes storing answers saved by logging into GOV.UK One Login in the forms-runner database


chao-xian (approved):

ADR/ADR054-store-saved-answers-in-the-database.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Decision
+
+When a form filler saves their form progress, we will save a snapshot of their answers to the forms-runner PostgreSQL database. We will link the data to the user's One Login account using the `sub` property returned by the One Login API, which is a unique identifier that does not change.
+
+When the user continues answering questions after they've saved their progress or returned to the form, we will continue to store their answers in their session in Redis until they save their progress again.
+
+We will encrypt the database column that contains the answers at the Active Record level, as we do for the `submissions` table.
+
+We will store the user's email address against their saved answers so we can identify the data to delete if a user contacts us to request deletion. We will encrypt this using deterministic encryption, which still allow us to query records using the email.
```
theseanything: Just to be clear that it's not the actual email address:
```suggestion
We will store a hash of a user's email address against their saved answers so we can identify the data to delete if a user contacts us to request deletion. We will encrypt this using deterministic encryption, which still allow us to query records using the email.
```

ADR/ADR054-store-saved-answers-in-the-database.md
```diff
@@ -0,0 +1,34 @@
...
+Records will be deleted:
+- when the user submits the form
+- 30 days after the saved progress was last updated
+
+## Consequences
+
+- We will be storing user data for a longer time, including forms that are never submitted.
+- Developers with access to the support role in the production environment will be able to access answers for incomplete forms if we add a rake task that decrypts the data. They already have access to submitted form data via a rake task that is used only when data is malformed.
+- Developers with access to the admin role in production will have access to the encryption key to allow them to decrypt answers for incomplete forms.
+- We will store identifiable information (the user's email address) alongside their answers if they use save and return. This will be encrypted and will not be logged.
```
theseanything:
```suggestion
- We will store identifiable information (hash of the user's email address) alongside their answers if they use save and return. This will be encrypted and will not be logged.
```

theseanything (dismissed):

ADR/ADR054-store-saved-answers-in-the-database.md
stephencdaly: Ah right yeh, a hash is probably better than encrypting if we only need it for querying and never want to be able to read it. We don't also need to encrypt the email address then.

stephencdaly: I've updated to say we'll store a hash

DavidBiddle (approved):

